Issue #148, opened 2013-08-10
reenable gravatar fallback


ralphbean:
It used to be that we would first check libravatar for an image, if that failed we'd fallback to gravatar, and if that failed we'd fall back to the fedora logo.

One week before flock, I broke that (commented it out) because I realised that using OpenID as the key for libravatar was _way better_ than using USER@fedoraproject.org as the key.

That gravatar fallback code should probably be re-added.  We'll have to still use http://USERNAME.id.fedoraproject.org for libravatar and USERNAME@fedoraproject.org for gravatar.  Tricky, not impossible.

PR #209, opened 2013-08-30
Reenable gravatar fallback

hroncok:
Fixes issue #148


ralphbean:
We're still blocking on upstream here.

I'm going to close this pull request, only because it clutters a pull request queue that we want to try and keep clean.  The code will still be around in @hroncok's branch and we can look forwards to re-opening it when we get progress on the libravatar side of things.

PR #290, opened 2015-02-11
Add a history pane that shows what badges were awarded for.

ralphbean:


ralphbean:
It looks like this.  The links take you to fedmsg messages in datagrepper.

tahrir

We've been adding those links to the db for a long time, but just not showing them in the UI.  https://github.com/fedora-infra/fedbadges/blob/develop/fedbadges/consumers.py#L218-L219

lmacken:
:ok_hand:

PR #291, opened 2015-02-12
Use UTC when coming up with the relative time text.

### LINKED ISSUE #257: "4 hours from now" bug
I awarded 3 of the same badge to 3 different people within 1 minute on badges.rit.edu and the last two came up as being awarded "4 hours from now":
hours_from_now

ralphbean:
Fixes #257.


pypingou:
:+1:

PR #299, opened 2015-04-01
Make the history pane less unwieldy.

### LINKED ISSUE #295: The History activity is long
I think it would be good to show the last 10 activities and place a link That says "show all activities" to make it more friendly.

It would also be good to hide it when watching from the mobile

ralphbean:
3a34180 - Completely hide the history pane in the mobile view.
acb60ba - Shorten the number of entries in the user-history pane.

Fixes #295.


pypingou:
Cool :+1:

PR #300, opened 2015-04-02
Use css flexbox to render badge collections much more nicely.

ralphbean:
See https://css-tricks.com/snippets/css/a-guide-to-flexbox/


pypingou:
Cool :+1:

Issue #122, opened 2013-07-29
Allow different sorting methods for badges.


oddshocks:
This isn't urgent, but we could add a few buttons to the user view so that you can sort badges by id, name, or whatever. For now, I'm fixing #120 by sorting the badges by id.

ralphbean:
I'm going to call this closed now that #301 is merged, released, and deployed.  It's not multiple sorting methods, but it *is* a much more reasonable way to view existing badges.

PR #301, opened 2015-04-02
Lay out badges according to their tags.

ralphbean:
This relates to #122 and we might even say it fixes it.  It should make the
default view readable enough that we don't need all kinds of custom sorting
(which would be non-trivial to implement).

### DIFF
--- a/tahrir/templates/user.mak
+++ b/tahrir/templates/user.mak
@@ -208,7 +208,7 @@
         % if tag in sum([badge.tags.strip(",").split(",") for badge in user_badges], []):
         <h3 class="section-header">${tag.title()} Badges</h3>
         <div class="flex-container">
-          % for badge in [b  for b in user_badges if tag in b.tags]:
+          % for badge in [b for b in user_badges if tag in b.tags.strip(",").split(",")]:
             ${self.functions.badge_thumbnail_flex(badge, 128, 33)}
           % endfor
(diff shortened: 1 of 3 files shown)


tahrir/templates/user.mak
```diff
@@ -204,11 +204,24 @@
             (<strong>${"{0:.1f}".format(percent_earned)}%</strong> of total).
           % endif
         </div>
+        % for tag in request.registry.settings.get('tahrir.display_tags', '').split(","):
+        % if tag in sum([badge.tags.strip(",").split(",") for badge in user_badges], []):
+        <h3 class="section-header">${tag.title()} Badges</h3>
         <div class="flex-container">
-          % for i, badge in enumerate(user_badges):
+          % for badge in [b  for b in user_badges if tag in b.tags]:
```
pypingou: Is ``b.tags`` here a list or a string? If it is a string that might cause problem if we have: event and non-event tags

ralphbean: It is a comma-separated string.  I'll update it to do b.tag.split(",").

ralphbean:
OK.  d46cfe5 addresses your comments, @pypingou.

pypingou:
:+1: I'm not sure I caught everything but let's give it a try on stg and see if it breaks somewhere :)

PR #302, opened 2015-04-02
Add an admin interface for adding new tags to existing badges.

ralphbean:


tahrir/templates/badge-base.mak
```diff
@@ -44,6 +44,28 @@
...
+                required
+                style="width: 100%;" />
+            </div>
+            <div class="grid-50">
+              <input class="pretty-submit"
+                style="height: 30px; width:100%"
+                name="add"
+                type="submit"
+                value="Add" />
+            </div>
```
pypingou: No csrf protection?

pypingou:
One question, otherwise :+1: 

Btw, can tag be removed?

ralphbean:
> Btw, can tag be removed?

Nope.  :)  You'd have to edit the db by hand.  (Before this PR, you'd have to edit the db by hand in order to add a tag also, so.. progress.)

pypingou:
On Thu, Apr 02, 2015 at 10:18:28AM -0700, Ralph Bean wrote:
>      Btw, can tag be removed?
> 
>    Nope. :) You'd have to edit the db by hand. (Before this PR, you'd have to
>    edit the db by hand in order to add a tag also, so.. progress.)

One approach I took on pagure is that:
the field contains all the current value:
- add one, it gets added
- remove one, it gets removed

Logic is here:
http://pagure.dev.fedoraproject.org/pagure/blob/master/pagure/lib/__init__.py#_1353

PR #304, opened 2015-04-06
Add fedmenu.

ralphbean:


development.ini.example
```diff
@@ -71,6 +71,13 @@ tahrir.use_openid_email = False
 
 tahrir.use_websockets = True
 
+# If you leave fedmenu.url undefined, it won't be written to the page at all.
+# https://github.com/fedora-infra/fedmenu
+#fedmenu.url = https://apps.fedoraproject.org/fedmenu
+#fedmenu.data_url = https://apps.fedoraproject.org/js/data.js
+fedmenu.url = http://threebean.org/fedmenu
+fedmenu.data_url = http://threebean.org/fedmenu/dev-data.js
```
pypingou: Should the example point to the prod instance instead of threebean.org?

pypingou:
One question, otherwise :+1:

development.ini.example
ralphbean: No, because it's not working in prod yet.  That requires a freeze break requests since I need to push content out to the prod proxies, and this isn't worth that.  ;)  I'll keep the current demo up at threebean.org so this will continue to work.

PR #306, opened 2015-04-24
Don't mangle the name for new badges passed in by the admin form.

ralphbean:


tahrir/views.py
```diff
@@ -201,16 +201,17 @@ def admin(request):
             else:
                 request.session.flash("Person with email {0} already exists.".format(email))
         elif request.POST.get('add-badge'):
-            idx = badge_name_to_id(request.POST.get('badge-name'))
+            name = request.POST.get('badge-name')
+            idx = badge_name_to_id(name)
             if not request.db.badge_exists(idx):
                 # Add a Badge to the DB.
-                request.db.add_badge(idx,
+                request.db.add_badge(name,
```
pypingou: We go from an index to a name here?

ralphbean: It was always a name.. but it got mistakenly changed to an ``id`` in a recent release.  This changes it back to be correct.

pypingou: ok

pypingou:
:+1: then

tahrir/views.py
ralphbean: See the list of released badges here for the mangled names:

https://badges.fedoraproject.org/explore/badges/rss

It starts around February, 2015.

PR #307, opened 2015-04-24
Allow users to specify a limit to the json leaderboard.

ralphbean:
With this, you could then set limit=1000000 to get a json response
containing all users in the db if you wanted to, say, do analytics on
Fedora Badges for a school project.

### DIFF
--- a/tahrir/views.py
+++ b/tahrir/views.py
@@ -460,6 +460,7 @@ def leaderboard(request):
 def leaderboard_json(request):
     """ Render a top-users JSON dump. """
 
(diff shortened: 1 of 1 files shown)


pypingou:
Could this lead to some sort of DDOS attack?

ralphbean:
Well, we were already calculating the entire leaderboard anyways (that's the compute/IO intensive part).  This just returns the data we already had in memory to the user.

pypingou:
ok :+1: then

PR #322, opened 2015-11-03
Changes for RPM-based distributions

trishnaguha:
For RPM-based distributions it should be /usr/bin/virtualwrapper.sh
Since there is no such file named as development.ini.sample it can be changed to development.ini.example, same for secret.ini.sample


ralphbean:
Looks good to me.  Thanks @trishnaguha!

trishnaguha:
Thanks @ralphbean :)

PR #323, opened 2015-11-04
Adding requirements to avoid traceback from dependencies

trishnaguha:
I got traceback from dependencies while hacking tahrir. So requirements.txt may fix it.


pypingou:
This is the list of everything installed in your virtualenv, what you want to list in your ``requirements.txt`` is the list of requirements of the project itself.

So if tahrir depends on ``pyramid-mako`` and installing ``pyramid-mako`` on ``mako`` itself, then you do not want to list ``mako``, it will be dragged while installing what you actually depend on which is ``pyramid-mako``

PR #324, opened 2015-11-04
Changes to fix traceback from dependencies

trishnaguha:
Adding requests and setting an upper bound to rdflib to fix traceback from dependencies.
Thanks @ralphbean and @pypingou


pypingou:
:+1: for me, this fixed it

ralphbean:
Thanks both!

PR #327, opened 2015-12-01
Fix the monthly report, same breakage

puiterwijk:
Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>

### DIFF
--- a/tahrir/views.py
+++ b/tahrir/views.py
@@ -1338,7 +1338,10 @@ def report_year_month(request):
     month = int(request.matchdict.get('month'))
 
     start = date(year, month, 1)
-    stop = date(year, month + 1, 1) - timedelta(days=1)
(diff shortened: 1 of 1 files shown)


puiterwijk:
Note: this fix is only valid for countries that start their years on January 1st, from the year they started doing that.
Since Tahrir was never used before every country that uses the Gregorian calendar did this (and since python datetime assumes Gregorian calendar, Tahir assumes that), this shouldn't be an issue, but just pointing out the limitation.

My other idea was to instead just do a date(year, month, 1) + timedelta(days=X), but that would require determining the number of days in a particular month, and if that's preferred I can update accordingly.

pypingou:
:+1: as well

puiterwijk:
Thanks!

PR #328, opened 2015-12-05
fix the percentile for the top users in the leaderboard

sayanchowdhury:
Currently the top users in the leaderboard see (top 0.0%) instead it should be something like (top 0.01%)


ralphbean:
Nice!  :+1:

PR #333, opened 2016-02-29
Linking GoBack of datagrepper page to user page

### LINKED ISSUE #332: "Go Back" on "this activity" page is not working
Clicking on `this activity` of a badge earned from `History` by default opens the raw data/datagrepper page in a new tab, when I click on the option `Go Back` of the datagrepper page it doesn't take me back to `user` page. 

Suppose for my username, https://badges.fedoraproject.org/user/trishnag when I click on `this activity` of one of the badges earned, the datagrepper page opens in a new tab. And when I click on `Go Back` of the datagrepper page, it stays on the same page. It should take me back to https://badges.fedoraproject.org/user/trishnag

So we can avoid `this activity` page i.e, the `datagrepper` data opening by default in a new tab. This might fix the issue?

trishnaguha:
Fixes #332

### DIFF
--- a/tahrir/templates/user.mak
+++ b/tahrir/templates/user.mak
@@ -165,7 +165,7 @@
             ${self.functions.badge_thumbnail(assertion.badge, 64, 33)}
             <div class="grid-66 text-64"><p>received on ${assertion.issued_on.strftime('%Y-%m-%d')}
               % if assertion.issued_for:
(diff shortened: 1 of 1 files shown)


ralphbean:
Sure, :+1:.  Thanks @trishnaguha!

PR #335, opened 2016-03-02
Change "add assertion" to "award individual badge"

cydrobolt:
"add assertion" sounds very ambiguous and difficult to understand, casting doubt about the button's action

If there is a specific reason why it needs to be "add assertion", please enlighten me :)

### DIFF
--- a/tahrir/templates/admin.mak
+++ b/tahrir/templates/admin.mak
@@ -226,7 +226,7 @@
         </tr>
         <tr>
           <td></td><td>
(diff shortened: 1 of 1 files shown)


ralphbean:
:+1: to the idea.. can we make the text just ``Award Badge`` though?  It sounds more crisp to me.

cydrobolt:
We could,  but there is another "award badge" button for awarding from a csv,  so we should probably distinguish from the two somehow.

ralphbean:
Ah, ok.  :+1: as is then.

PR #344, opened 2016-06-21
Create a function to return a dict of user badge informations

sayanchowdhury:


tahrir/views.py
```diff
@@ -70,6 +70,47 @@ def _get_user(request, id_or_nickname):
...
 
+def _get_user_badge_info(request, user):
+    """ Returns a dictionary of the user badge information
+    """
+
+    # Get user badges.
+    user_badges = [a.badge for a in user.assertions]
+
+    # Sort user badges by id.
+    user_badges = sorted(user_badges, key=lambda badge: badge.id)
```
pypingou: Why the two lines here? We could do something like ``user_badges = sorted([a.badge.id for a in user.assertions])`` no?

/me always try to avoid lambda function when possible

pypingou:
One comment, otherwise looks good

tahrir/views.py
sayanchowdhury: `user_badges = sorted([a.badge.id for a in user.assertions])` would just return the list of sorted badge ids.

pypingou: Ah I see, this is sorted the badge by their id, ok then :)

pypingou:
:+_

sayanchowdhury:
Thanks :)

tahrir/views.py
ralphbean: FWIW, if you really wanted to avoid the lambda you could use:

```python
import operator
user_badges = sorted(user_badges, key=operator.attrgetter('id'))
```

The lambda definitely works!

Issue #343, opened 2016-04-21
RFE: Badge paths


mairin:
A commonly requested feature is having a formal sequence representation for badges in a series. The sequence should have a particular order. Examples of badge series that exist today:
- Badge artist series (ex https://badges.fedoraproject.org/badge/apprentice-badge-artist-i)
- Badge milestone series (ex https://badges.fedoraproject.org/badge/baby-badger)
- Tester series (ex. https://badges.fedoraproject.org/badge/in-search-of-the-bull-tester-i) 

We need a way to store:
- What series is this badge a part of
- What badge came before it, if any
- What badge comes after it, if any
- (Or simply which order # is the badge in the series, where the first is 1 and the second is 2, so on and so forth.)

This can help drive onboarding, especially where teams have mapped out badge paths to help a new contributor gain experience.

Related ideas along this line have been requested a few times, see:
https://github.com/fedora-infra/tahrir/issues/270
https://github.com/fedora-infra/tahrir/issues/204
https://github.com/fedora-infra/tahrir/issues/250

Some of the other RFEs request specifying how many {action}s you have to take to get the next badge. I don't think that is necessary and it makes it a harder problem to solve. I think simply storing what badge comes next, and having a section of the UI when you log in that displays the 'next' badges for the ones you've already earned so you know what badges you're closest to earning next would go a long way and would essentially solve the same problem.

We'd like to use badge paths in hubs too, and I think it makes the most sense to implement them in badges and display in hubs.

mairin:
Some things I don't understand and would like clarity on:

https://git.fedorahosted.org/cgit/badges.git/tree/rules/ambassadors-sponsor.yml

So each badge has a list of attributes at the front. Is that where we'd put this? Can the attribute be empty or non-present? So we'd have these new attributes with example values(suggesting the name 'series' because 'path' might be overloaded):

series:Badger Artist
series_order: 1

PR #345, opened 2016-06-26
Add series information to badge page

lubomir:
Add a box to the badge detail page that shows all badges in the same series. The relationship is determined based on badge name. Current badge is highlighted with blue-tinted background.

The assumption on name is that all badges in a series will have a parenthesised series name and a number in the name. The number can be either in Arabic or Roman numerals.

This holds for almost all current badges in Fedora. The only exceptions are:

 * Free the Fedora: has numbers, but not in parentheses
 * Top: Gold, Silver and Bronze are not tied the the others; badges with numbers actually work in the opposite direction (Top 10 is after Top 500 in the series)
 * Editor and Corporate Ant series have no indication that the badges belong together

snimek z 2016-06-26 19-51-14

### DIFF
--- a/tahrir/static/css/tahrir.css
+++ b/tahrir/static/css/tahrir.css
@@ -429,3 +429,8 @@ footer > .document {
 .header > h1 {
     display: inline-block;
 }
+
+.current-badge .thumbnail-container {
+    background: #dfdfef;
+    border-radius: 50%;
+}
--- a/tahrir/templates/badge-base.mak
+++ b/tahrir/templates/badge-base.mak
@@ -165,6 +165,24 @@
         </ul>
       </div> <!-- End padded content. -->
     </div> <!-- End shadow. -->
+    % if related_badges:
+      <div class="shadow">
+        <h1 class="section-header">Series <em>${series_name}</em></h1>
+        <div class="padded-content">
+          <div class="flex-container">
+            % for b in related_badges:
+              % if b[1] == current_idx:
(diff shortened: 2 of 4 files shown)


ralphbean:
:boom:

sayanchowdhury:
Cool :+1:

lubomir:
Actually, there are more cases where this does not detect series correctly. I tested on a checkout of https://git.fedorahosted.org/git/badges.git which does not seem to have all badges that are currently in production.

The frog series for having FAS account for X years should definitely form a series, but does not.

These badges might make sense a series, but are not detected:
* Nuancier
* Badges for proposing supplementary wallpapers
* Having them accepted
* Having a Change accepted

ralphbean:
Maybe you could supplement the automatic detection with a list of series defined in the configuration file, perhaps?

A more permanent solution (while more difficult) would involve modifying the database to definitively link badges together in a series.

sayanchowdhury:
Yes, that's the think I was thinking about too as I started to work on [this](https://github.com/fedora-infra/tahrir/issues/343) which is one the same lines

lubomir:
My initial idea was to do this without having to modify the data storage. However, the amount of cases where this heuristic fails is rather large, so storing the data explicitly probably makes more sense.

https://github.com/fedora-infra/tahrir/pull/346
https://github.com/fedora-infra/tahrir-api/pull/37

Is there some UI for editing the badges?

PR #347, opened 2016-08-03
Forms to create Teams, Series, Perks and json endpoint for the Paths

sayanchowdhury:


tahrir/templates/admin.mak
```diff
@@ -83,6 +83,100 @@
...
+        </tr>
+      </table>
+    </form>
+  </div>
+
+  <div class="clear"></div>
+
+  <div class="grid-33">
+    <h2>Perk</h2>
+    <p>Add a new Perk</p>
```
pypingou: Perk isn't quite a word I'm used to or know well, should we add a one-liner description of what perks are?

tahrir/views.py
```diff
@@ -70,6 +63,15 @@ def _get_user(request, id_or_nickname):
             return None
 
 
+def _get_team(request, team_id):
+    '''Get team for the given team_id'''
+    team = request.db.get_team(team_id=team_id)
+
+    if team:
+        return team
+    return None
```
pypingou: What values could team have? A team object or ?
Just curious in which case ``if team`` will return `False` while the team object isn't a `None` already :)

pypingou:
A couple of comments but looks alright to me otherwise

tahrir/templates/admin.mak
```diff
@@ -83,6 +83,100 @@
...
+
+  <div class="clear"></div>
+
+  <div class="grid-33">
+    <h2>Series</h2>
+    <p>Add a new Series</p>
+  </div>
+
+  <div class="grid-66">
+    <form class="admin-form" method="POST">
```
pypingou: Same here (and below)

tahrir/views.py
pypingou: Could the db return you things like ``''`` or ``False`` ? If not, then I think returning team directly is just as good :)

tahrir/views.py
```diff
@@ -244,7 +296,7 @@ def admin(request):
                 request.session.flash("Person with email {0} already exists.".format(email))
         elif request.POST.get('add-badge'):
             name = request.POST.get('badge-name')
-            idx = badge_name_to_id(name)
+            idx = convert_name_to_id(name)
```
pypingou: Did we check if this method is called elsewhere?

pypingou:
A few comments, mainly minor things. The only points I would discuss is the use of the term ``perk`` and I think I like the idea of ``quest`` better than the one of ``path`` (which I know I introduced, but I think quest conveys a nicer idea, especially in the badges world)

tahrir/templates/admin.mak
```diff
@@ -83,6 +83,100 @@
...
 
+  <!-- TEAM PANEL -->
+
+  <div class="grid-33">
+    <h2>Team</h2>
+    <p>Add a new team</p>
+  </div>
+
+  <div class="grid-66">
+    <form class="admin-form" method="POST">
```
sayanchowdhury: Sends request to the same endpoint.

tahrir/views.py
sayanchowdhury: Yes, I did check for the other locations for this method.

sayanchowdhury:
`Perk` is term used when describing each point in a `skill tree`. `Quest` sounds to me better than `Path` too.

tahrir/views.py
sayanchowdhury: `get_team` returns the `team object` or `None`. I guess I can return just the team.

tahrir/templates/admin.mak
pypingou: just personnal preferences, but I prefer to be explicit :)

sayanchowdhury:
@pypingou Updated the PR. Looks good now?

tahrir/templates/admin.mak
```diff
@@ -83,6 +83,100 @@
...
+    <p>Add a new team</p>
+  </div>
+
+  <div class="grid-66">
+    <form class="admin-form" method="POST" action="${request.route_url('admin')}">
+      <input type="hidden" name="csrf_token" value="${request.session.get_csrf_token()}"/>
+      <table>
+        <tr>
+          <td><label for="team-name">Team Name</label></td>
+          <td><input name="team-name" required="required" placeholder="Creative Team"/></td>
```
pypingou: `placeholder="Creative Team"` ?

tahrir/templates/admin.mak
```diff
@@ -83,6 +83,100 @@
...
+        </tr>
+      </table>
+    </form>
+  </div>
+
+  <div class="clear"></div>
+
+  <div class="grid-33">
+    <h2>Series</h2>
+    <p>Add a new Series</p>
```
pypingou: -s no?

tahrir/views.py
```diff
@@ -227,7 +227,57 @@ def admin(request):
...
+            else:
+                request.session.flash(
+                    "Series with name {0} already exists.".format(series_name))
+
+        elif request.POST.get('add-milestone'):
+            data = request.POST
+            series_id = data.get('milestone-series-id')
+            badge_id = data.get('milestone-badge-id')
+            if not request.db.milestone_exists_for_badge_series(badge_id,
+                                                           series_id):
```
pypingou: You may want to split after the ``(``, with 1 indentation more than the logic bloc it should be valid pep8

tahrir/views.py
```diff
@@ -227,7 +227,57 @@ def admin(request):
...
+                position = data.get('milestone-position')
+                request.db.create_milestone(position=position,
+                                       series_id=series_id,
+                                       badge_id=badge_id)
```
pypingou: Either same as above, or it needs one more indentation :)

pypingou:
Looks good to me, feel free to do a small pep8 pass (if you like) and merge :)

tahrir/templates/admin.mak
sayanchowdhury: That's an example name  for the team:)

tahrir/templates/admin.mak
sayanchowdhury: You mean Serie?

sayanchowdhury:
There are couple of PEP8 fixes needed in both tahrir and tahrir_api. Fixing them right now

tahrir/templates/admin.mak
pypingou: yup

PR #353, opened 2016-12-01
Fix breaking bugs

puiterwijk:
Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>

### DIFF
--- a/tahrir/templates/user.mak
+++ b/tahrir/templates/user.mak
@@ -31,7 +31,7 @@
 
           % if rank != 0:
             <p>
-              Ranked ${rank} out of ${user_count} ranked users (top ${"{0:.2f}".format(percentile)}%).</p>
+              Ranked ${rank} out of ${user_count} ranked users (top ${"{0:.2f}".format(float(percentile))}%).</p>
           % else:
             <p>Not ranked yet.</p>
           % endif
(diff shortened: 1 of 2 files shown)


tahrir/views.py
```diff
@@ -1437,6 +1437,7 @@ def report_year_month(request):
     year = int(request.matchdict.get('year'))
     month = int(request.matchdict.get('month'))
 
+    now = datetime.now()
```
pypingou: utcnow?

puiterwijk: It is only using the date part of it, so not sure it matters... Also, our servers are all on UTC.

tahrir/templates/user.mak
```diff
@@ -31,7 +31,7 @@
 
           % if rank != 0:
             <p>
-              Ranked ${rank} out of ${user_count} ranked users (top ${"{0:.2f}".format(percentile)}%).</p>
+              Ranked ${rank} out of ${user_count} ranked users (top ${"{:0.2f}".format(float(percentile))}%).</p>
```
sayanchowdhury: What's the issue here?

tahrir/views.py
sayanchowdhury: +1 for this one. I thought I pushed this fix earlier

pypingou: I'd still prefer we stick to UTC as not everyone runs their server with UTC :)

PR #362, opened 2017-03-08
Added Vagrant setup for hacking on Tahrir

ryanlerch:


README.rst
```diff
@@ -133,6 +133,37 @@ and append ``YOUR_FAS_USERNAME@fedoraproject.org`` to the ``tahrir.admin`` optio
...
+   $ cp Vagrantfile.example Vagrantfile
+
+Run the ``vagrant up`` command to provision your dev environment::
+
+    $ vagrant up
+
+When this command is completed (it may take a while) start tahrir with the
+following command:
+
+    $ vagrant ssh -c"cd /vagrant/; pserve --reload development.ini"
```
bowlofeggs: I recently converted the Bodhi Vagrant environment to put a systemd unit file on the box so that Bodhi is automatically run for you. It's pretty sweet, and it means the user doesn't have to do the above step. Here's some links to the bits that do this:

https://github.com/fedora-infra/bodhi/blob/2.4.0/devel/ansible/roles/dev/files/bodhi.service
https://github.com/fedora-infra/bodhi/blob/2.4.0/devel/ansible/roles/dev/tasks/main.yml#L103-L107
https://github.com/fedora-infra/bodhi/blob/2.4.0/devel/ansible/roles/dev/tasks/main.yml#L123-L127

It's just a suggestion, but I think it's a nice finishing touch.

sayanchowdhury:
I am going ahead an merging the changes and then apply the other requests on top of this.

sayanchowdhury:
Thanks @ryanlerch :+1:

PR #365, opened 2017-04-11
Fix the json output for series

sayanchowdhury:
Right now only the last series is returned in the json output.
This fixes the issues and returned all the series in the output
This change also returns the milestone information in different
lists for the convenience.

Signed-off-by: Sayan Chowdhury <sayan.chowdhury2012@gmail.com>

### DIFF
--- a/tahrir/views.py
+++ b/tahrir/views.py
@@ -1167,16 +1167,17 @@ def _user_team_json_generator(request, team, user):
     assertions = user.assertions
     assertion_ids = set([assertion.badge_id for assertion in assertions])
 
+    series_info = []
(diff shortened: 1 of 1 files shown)


pypingou:
Is there any risk this could dump too much data? (in other words, should this be paginated?)

sayanchowdhury:
I don't think so atm we would need to paginate. But, I can start working on this after merging the PR. Right now at one go there would be 20-30 badges information

pypingou:
Code wise 👍 for me

sayanchowdhury:
Thanks :+1:

PR #377, opened 2018-01-16
Don't calculate badge stats when getting user badges

puiterwijk:
For people with many badges, this brings down the user json generation
from >50 seconds to <4 seconds.

Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

### DIFF
--- a/tahrir/views.py
+++ b/tahrir/views.py
@@ -819,7 +819,16 @@ def badge(request):
             )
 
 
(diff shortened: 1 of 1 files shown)


relrod:
As discussed on IRC, this breaks API compat, but people shouldn't be pulling that data from this endpoint anyway, so :+1:

PR #384, opened 2018-04-05
Exorcise moksha.wsgi.

ralphbean:
The backend bits of moksha are py3 ready, but the moksha.wsgi piece is not, plus it has some hairy deps that still need to be ported:

http://fedora.portingdb.xyz/pkg/python-moksha-wsgi/

Let's retire it.  Removing it here to facilitate that.

### DIFF
--- a/development.ini.example
+++ b/development.ini.example
@@ -5,15 +5,11 @@ port = 8000
 
 [pipeline:main]
 pipeline =
-    moksha
     tw2
     pyramid
 
 [filter:tw2]
 use = egg:tw2.core#middleware
-
-[filter:moksha]
-use = egg:moksha.wsgi#middleware
 
 [app:pyramid]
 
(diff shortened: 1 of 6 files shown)


ralphbean:
If we can merge this, I'll retire `python-moksha-wsgi`.

ralphbean:
@pypingou do you know if anyone is looking after `tahrir` these days?

I'm trying to resolve some of the py3-only issues in Fedora packages.

sayanchowdhury:
I look into `tahrir` these days

sayanchowdhury (approved):

ralphbean:
Awesome!  Thanks @sayanchowdhury.

I'll plan to remove python-moksha-wsgi from rawhide for http://fedora.portingdb.xyz/pkg/python-moksha-wsgi/

ralphbean:
Wait... I can't do that.  The `fedora-packages` app still depends on python-moksha-wsgi.  :(

I guess.. I could retire it in rawhide and we could still carry it in epel7-infra, but that's a pain.

Will have another look at porting it instead.

PR #391, opened 2019-01-21
Tox

bowlofeggs:
The primary goal of this pull request is to add Tox as a test runner. There are three commits. The first two fix some Sphinx issues and add a unit test (so Tox has a unit test to run). The third adds tox and configures it to run the tests, enforce diff coverage, and build the docs. It also adds a ```.cico.pipeline``` file so that we can have pull request CI.

### DIFF
--- a/.cico.pipeline
+++ b/.cico.pipeline
@@ -1,0 +1,1 @@
+fedoraInfraTox { }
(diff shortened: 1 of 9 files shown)


cverna (approved):
LGTM

PR #395, opened 2019-01-31
Drop support for foaf format.

cverna:
Foaf format is not very popular and dropping support
removes quite a bit of code from tahrir

Signed-off-by: Clement Verna <cverna@tutanota.com>


sayanchowdhury:
:100:

sayanchowdhury (approved):

sayanchowdhury (approved):

PR #427, opened 2019-08-09
Initial Python 3 port

AdamWill:
These changes together make Tahrir work at least to some extent on Python 3, for me, using latest git tahrir-api with [this PR](https://github.com/fedora-infra/tahrir-api/pull/47). Everything I poked and tested so far works, of course the test suite is basically non-existent so it's hard to know what else lurks. I at least tested creating badges, awarding them, poking around the web UI a bit, loading JSON and RSS feeds and stuff.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ requires = [
     "requests",
     "rdflib<=4.1.2",
     "pytz",
+    "six",
 
     # For qrcode to work from PyPI, you also need Pillow.
     # This is handled for us in Fedora because python-qrcode pulls in the
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,6 @@ requires = [
     'transaction',
     'pyramid_tm',
     'zope.sqlalchemy',
-    "weberror",
     'velruse',
     "qrcode",
     "dogpile.cache",
--- a/setup.py
+++ b/setup.py
@@ -18,9 +18,10 @@ requires = [
     "dogpile.cache",
     'docutils',
     "python-dateutil",
-    "webhelpers",
+    "feedgen",
     "requests",
     "rdflib<=4.1.2",
+    "pytz",
 
     # For qrcode to work from PyPI, you also need Pillow.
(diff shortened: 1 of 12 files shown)


AdamWill:
@christophedumont5 I'm guessing you're the Chistophe Dumont from infra list? :)

AdamWill:
CI failed because coverage of the diff is <100%. I'd have to write quite a lot of tests to get 100%. I might write some, but I'm not promising it. :)

mscherer:
look good to me

laxathom:
> These changes together make Tahrir work at least to some extent on Python 3, for me, using latest git tahrir-api with [this PR](https://github.com/fedora-infra/tahrir-api/pull/47). Everything I poked and tested so far works, of course the test suite is basically non-existent so it's hard to know what else lurks. I at least tested creating badges, awarding them, poking around the web UI a bit, loading JSON and RSS feeds and stuff.

Eh, right :smiley: . Found an issue on route `/report` which leads leaderboard page to 500-ing.
You also need to cast that `dict_items()` to a `list` [here.](https://github.com/fedora-infra/tahrir/blob/develop/tahrir/templates/report.mak#L9)

Other than this, I played a little w/ only one badge and one user, then w/o any data. Looks good so far.
Might also need to test it w/ more data like a 1000 badges. That something we can test after opened the release branch.

AdamWill:
nice catch, done (though hmm, this thing actually looks fairly slow and inefficient....but changing that would be beyond the scope of a python3 port...)

laxathom:
> nice catch, done (though hmm, this thing actually looks fairly slow and inefficient....but changing that would be beyond the scope of a python3 port...)

Correct. I thought about performances but it would take more time to factor this. We better make it work first then improve.
Most of the code will drastically change according to what the design team wants to go with the user experience.

laxathom (approved):

cverna:
@laxathom and @AdamWill thanks

PR #446, opened 2021-07-22
Updating with the new Fedora logo

mairin:
This addresses issue https://pagure.io/fedora-design/spot-the-old-logo/issue/20


mscherer:
So, we are in the (slow) process of replacing tahrir, and I am not even sure how we can easily deploy a new version. So I am ok to merge, but I feel that deploying the new logo might open a can of worm.

Issue #370, opened 2017-08-20
404 page should not include JQuery


QuLogic:
This is the breakdown of some random non-existent page on badges.fp.o, according to Firefox (primed cache is the same):

![screenshot from 2017-08-19 23-37-06](https://user-images.githubusercontent.com/302469/29491874-5d57ca4e-8537-11e7-8f09-0294e8bf827e.png)

This is partially because of #369, but if JQuery were not included at all, the 404 page would load much faster as most of this time (for the js at least) is spent doing transfers:
![screenshot from 2017-08-19 23-40-56](https://user-images.githubusercontent.com/302469/29491888-e8cc8d80-8537-11e7-84fe-2d8d785a13ed.png)

PR #541, opened 2024-08-16
Fix the error pages

### LINKED ISSUE #352: 404/500 errors
404 and 500 error pages should have actual text, instead of just a centered image. The image is fine for the 404 or 500 badgers, but not for the text itself.

abompard:
- don't put the text inside the image
- add alt text
- use a single error page for all the error codes

Fixes: #352


tahrir/templates/error.html
```diff
@@ -0,0 +1,18 @@
+{% extends '_base.html' %}
```
QuLogic, line 1: Note that this might regress on #370, though that issue was never explicitly closed (so I don't know that it was ever intentionally fixed) and weirdly, I can't find jQuery anywhere in the history of `404.html`, so I'm not sure where it was coming from before.

abompard: Yeah we should probably try to get rid of jQuery entirely at some point.

PR #684, opened 2025-08-27
Move to React Bootstrap implementation

### LINKED ISSUE #664: Create the UI for the badges listing page
Create the UI for the badges listing page

gridhead:
Move to React Bootstrap implementation


gridhead:
To do hell with Material UI - All my homies like React Bootstrap...

image

image

image

image

image

image

laxathom (approved):
Looks good to me.

mscherer:
Shouldn't some stuff be adjusted for renovate to take care of the jsx code, etc ? 

And also I think a CI / linter would be a good idea for that new part of the code.

gridhead:
@laxathom thanks for your vote of confidence, but this PR is far from ready. There is a bunch more that needs to be done that we are merging into the [`swatantrya`](https://github.com/fedora-infra/tahrir/tree/swatantrya) branch as we go on. Once the revamp is ready, this pull request would be changed into the ready state. Then, the entire revamp could be reviewed for improvements or modifications.

@mscherer hello 👋🏼, absolutely! It's just that we have not gotten to that part yet.

gridhead:
Here are the badges listing page.

For complete collection, currently accessible through https://badges.gridhead.net/discover/recently.

image

For the recently included, currently accessible through https://badges.gridhead.net/discover/accolade.

image

@mscherer @laxathom @abompard @sdglitched please feel free to play around with this. If it seems to be broken at some point in time, just come back later as changes are actively made here.

frontend/src/components/modewrap.jsx
```diff
@@ -0,0 +1,21 @@
...
+import { useSelector } from "react-redux";
+
+export default function ModeWrap({ children }) {
+  const mode = useSelector((data) => data.area.mode);
+
+  useEffect(() => {
+    const updateMode = () => {
+      if (mode === "auto") {
+        const mode = window.matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "light";
+        document.body.setAttribute("data-bs-theme", mode);
```
abompard: The Bootstrap docs indicate that they expect the attribut to be set on the `<html>` element, not the `<body>` element (although it seems to work as well).
You can use `document.documentElement` to get that attribute.

frontend/src/components/modewrap.jsx
```diff
@@ -0,0 +1,21 @@
...
+      }
```
abompard, line 16: So, if `mode !== "auto"`, the  `updateMode()` function does nothing?

frontend/src/components/giveitem.jsx
```diff
@@ -0,0 +1,32 @@
+import { Link } from "react-router";
+import ListGroup from "react-bootstrap/ListGroup";
+import Badge from "react-bootstrap/Badge";
+import CryptoJS from "crypto-js";
+import { portraitProvider } from "../features/util.js";
+
+export default function GiveItem({ name, mail, rank, body }) {
+  return (
+    <ListGroup.Item
+      key={CryptoJS.SHA256(name).toString()}
```
abompard: Why the SHA256 here? If I remember correctly, `key` is only used to differentiate between items of a for loop, no? Why is it necessary here? And if it is, could `name` be sufficient?

frontend/src/components/category.jsx
```diff
@@ -0,0 +1,18 @@
+import Card from "react-bootstrap/Card";
+import Row from "react-bootstrap/Row";
+import CryptoJS from "crypto-js";
+
+export default function Category({ name, wide, children }) {
+  return (
+    <Card key={CryptoJS.SHA256(name).toString()} className="mb-2">
```
abompard: Same remark about the `key` attribute.

frontend/src/components/accolade.jsx
```diff
@@ -0,0 +1,139 @@
...
+          <Card.Body className="p-2">
+            <Card.Title className="dataelem text-truncate mb-1">{acco.name}</Card.Title>
+            <Card.Text className="small mb-2">{acco.description}</Card.Text>
+            <Stack direction="horizontal" gap={2}>
+              {acco.tags &&
+                acco.tags.map((name) => (
+                  <Badge
+                    as={Link}
+                    to={`/category/${name}`}
+                    key={CryptoJS.SHA256(name).toString()}
```
abompard: Why the SHA256? Just `name` should be sufficient here, no?

frontend/src/components/accoitem.jsx
```diff
@@ -0,0 +1,29 @@
...
+import Col from "react-bootstrap/Col";
+import { Link } from "react-router";
+import Image from "react-bootstrap/Image";
+
+export default function AccoItem({ iden, name, body, foot, shot }) {
+  return (
+    <OverlayTrigger
+      placement="auto"
+      overlay={
+        <Popover id={CryptoJS.SHA256(`over-${iden}`).toString()} className="bodyelem">
```
abompard: Why the SHA256? Wouldn't it be sufficient to use `over-${ident}`?

frontend/src/components/navigate.jsx
```diff
@@ -0,0 +1,72 @@
...
+            className="d-inline-block align-top p-0 logo"
+          />
+        </Navbar.Brand>
+        <Navbar.Toggle />
+        <Navbar.Collapse className="justify-content-end">
+          <Navbar.Text className="p-0">
+            <NavDropdown
+              title={
+                <Icon
+                  path={mode === "lite" ? mdiWeatherSunny : mode === "dark" ? mdiWeatherNight : mdiMemory}
```
abompard: The two modes supported by Bootstrap are `light` and `dark`, not `lite`.

Also, their documentation use the [circle-half-full](https://pictogrammers.com/library/mdi/icon/circle-half-full/) icon for the `auto` mode, I think it's clearer that the icon is about something related to color than the microprocessor icon.

frontend/src/components/navigate.jsx
```diff
@@ -0,0 +1,72 @@
...
+                className="small d-flex align-items-center p-1"
+              >
+                <Icon className="me-1" size={0.75} path={mdiMemory} />
+                System
+              </NavDropdown.Item>
+              <NavDropdown.Item
+                onClick={() => {
+                  dispatch(keepMode("lite"));
+                  document.body.setAttribute("data-bs-theme", "light");
+                }}
```
abompard: You can factor this handler in a separate function (if you keep the same name for the mode in Redux and the actual mode in the attribute)

abompard (commented):
Just a few comments I noticed, but I didn't review everything.

Are you sure it's a good idea to store `acco`, `user` and `list` in Redux? It looks like they're the result of backend queries only done for the purpose of displaying the current page, and they're not used across the app. It seems that the initial state in the store contains a lot of things that will have to be modified when the backend changes, that could be a source of future errors.

If you want to manage the result of REST queries, I would recommend using [react-query](https://tanstack.com/query/) or [RTK Query](https://redux-toolkit.js.org/tutorials/rtk-query)

Also, if I remember correctly, Redux recommends to separate the `components` that are reusable and don't depend on Redux from the `pages` that use redux and are thus not really reusable, but use components to display. I think it's useful advice.

abompard:
In the applications I've used Redux on, I ended up using it only for state that spans the entire application, such as the logged-in user, the color mode, the loading state, etc. So, not a lot.
All the backend-fetching state was handled by React Query (now TanStack) but I suspect RTK Query works the same.
All the form-specific state was handled by Formik.

gridhead:
@abompard I have made the suggested changes. Please feel free to take another look. Unless they are design related changes, please do not hold this PR up and merge it as soon as the architectural kinks that you highlighted are addressed.

frontend/src/components/giveitem.jsx
mscherer: And if hashing is needed, why sha256, wouldn't a shorter and faster hash be enough for that purpose (and be more friendly on people CPU and speed of rendering of the page ) ?

gridhead:
I have gone ahead to make some more advancements to the badge viewing page.

image

frontend/src/contexts/accolade.jsx
```diff
@@ -0,0 +1,136 @@
...
+        <Card className="mb-2">
+          <Card.Img variant="top" src={acco.image} />
+          <Card.Body className="p-2">
+            <Card.Title className="dataelem text-truncate mb-1">{acco.name}</Card.Title>
+            <Card.Text className="small mb-2">{acco.description}</Card.Text>
+            <Stack direction="horizontal" gap={2}>
+              {acco.tags &&
+                acco.tags.map((name) => (
+                  <Badge
+                    key={generateIdentity()}
```
abompard: This function seems to generate a random string each time, but that's not what the `key` attribute is for. The `key` is used by React to know if a sub-element must be re-rendered or not, here's [the docs about this](https://react.dev/learn/rendering-lists#rules-of-keys). Here, `name` is a good candidate: it's unique in the array and it won't change over time for this particular element.

frontend/src/contexts/accolade.jsx
```diff
@@ -0,0 +1,136 @@
...
+            </Stack>
+          </Card.Body>
+        </Card>
+        <Card className="mb-2">
+          <Card.Body className="ps-0 pe-0 pt-2 pb-0">
+            <Card.Title className="mb-0 ps-2 dataelem">First awarded</Card.Title>
+            <hr className="mt-2 mb-0" />
+            <ListGroup variant="flush">
+              <GiveItem
+                key={generateIdentity()}
```
abompard: Here you don't seem to be rendering an array, so you shouldn't need he `key` attribute.

frontend/src/contexts/accolade.jsx
```diff
@@ -0,0 +1,136 @@
...
+            </ListGroup>
+          </Card.Body>
+        </Card>
+        <Card className="mb-2">
+          <Card.Body className="ps-0 pe-0 pt-2 pb-0">
+            <Card.Title className="mb-0 ps-2 dataelem">Last awarded</Card.Title>
+            <hr className="mt-2 mb-0" />
+            <ListGroup variant="flush">
+              <GiveItem
+                key={generateIdentity()}
```
abompard: Same here, you shouldn't need it.

frontend/src/contexts/accolade.jsx
```diff
@@ -0,0 +1,136 @@
...
+            <Card.Title className="mb-0 ps-2 dataelem" style={{ textTransform: "capitalize" }}>
+              History
+            </Card.Title>
+            <Card.Text className="mb-0 ps-2 small">Awarded {acco.assertions.length} time(s)</Card.Text>
+            <hr className="mt-2 mb-0" />
+            <ListGroup variant="flush">
+              {acco.assertions &&
+                acco.assertions.map((assertion) => (
+                  <GiveItem
+                    key={generateIdentity()}
```
abompard: In this case, I think `assertion.name` is a good candidate, unless you have something like `assertion.id` in your object. There are other places where `generateIdentity()` is used, I'm going to save some space and not list them all in the review ;-)

abompard (commented):
A couple more comments!
The usage of `createApi` makes things much clearer I think, that's pretty cool :-)
Are you interested in comments about variable naming? ;-) Two main things come to my mind:
- the `context` directory doesn't seem clear to me at first, `context` has a rather vague meaning, if I understand correctly you chose that because the files there are *using* the Redux state? Since those files are basically pages (they are linked to directly by the `main.jsx` page in the `<Routes>` section, I think I would have used something like `pages` or `routes`.
- the `slugdata` variable is also very vague and could be more expressive. For example, in the `identity` page, it's a username. You could either use `username` in the `main.jsx` page or rename the variable in the `identity.jsx` page with `const { slugdata: username } = useParams();`

It's probably a bit early to talk about variable naming but since these affect quite a few pages I thought it'd be better to mention them earlier rather than later when there's more to refactor.

gridhead:
> The usage of `createApi` makes things much clearer I think, that's pretty cool :-)

Thanks!

> Are you interested in comments about variable naming? ;-) Two main things come to my mind:

No, I am not but since you've already provided... Oh well.

> - the `context` directory doesn't seem clear to me at first, `context` has a rather vague meaning, if I understand correctly you chose that because the files there are *using* the Redux state? Since those files are basically pages (they are linked to directly by the `main.jsx` page in the `<Routes>` section, I think I would have used something like `pages` or `routes`. 

That is what I had in mind but since you asked so nicely, `routes` it is.

> - the `slugdata` variable is also very vague and could be more expressive. For example, in the `identity` page, it's a username. You could either use `username` in the `main.jsx` page or rename the variable in the `identity.jsx` page with `const { slugdata: username } = useParams();`

It is a `data` that we are dealing with in the form of `slug` (URL safe - either dash separated, or underscore separated). I will retain the name `slugdata` in the `main.jsx` file but I will make the necessary changes in the specific routes to make the variable names, more specific. There would, of course, be a plenty of revisiting the names and types when we would be TypeScripting this.

Regarding the `generateIdentity` function, I insist that it stays. The keys are ensured to be of uniform length and uniform type, but I will remove the keys from wherever they need not be. I have faced warnings in the browser console for times when there was collision in the keys and also, sometimes the components do not get re-rendered if the key remains the same as before.

gridhead:
Very important

image

frontend/src/components/giveitem.jsx
gridhead: @mscherer I strongly suspect that it would be the network that would bottleneck the rendering and not the speed with which client-side JavaScript asynchronously generates SHA256 checksums.

gridhead:
@abompard please feel free to take another look.

abompard:
> Regarding the `generateIdentity` function, I insist that it stays. The keys are ensured to be of uniform length and uniform type

That's not a requirement for React's use of the `key` attribute if I remember correctly.

>, I have faced warnings in the browser console for times when there was collision in the keys

That means that the array contains duplicate objects, it which case it's good that React warns you about it. Unless you really want to render duplicate objects, in which case I suppose you can use the array index as the key.

> and also, sometimes the components do not get re-rendered if the key remains the same as before.

Then there must be another problem, but it's probably not because of the key. The key attribute is an important part of React's rendering optimization as explained in [this paragraph](https://react.dev/learn/rendering-lists#where-to-get-your-key). The docs also say:

> Similarly, do not generate keys on the fly, e.g. with key={Math.random()}. This will cause keys to never match up between renders, leading to all your components and DOM being recreated every time. Not only is this slow, but it will also lose any user input inside the list items. Instead, use a stable ID based on the data.

Could you show me a component that does not get re-rendered with a fixed key? Maybe I can help find the reason.

gridhead:
The keys are deterministic in nature now, check https://github.com/fedora-infra/tahrir/pull/684/commits/f68068325e0f7c973cc4c67fd14d6949224d8b4f.

I am merging this one.

frontend/src/features/util.js
```diff
@@ -0,0 +1,22 @@
...
+    timeZoneName: "short",
+  });
+}
+
+export function portraitProvider(mail, size) {
+  return `https://seccdn.libravatar.org/avatar/${md5(mail.trim().toLowerCase()).toString()}?s=${size}&d=retro`;
+}
+
+export function generateIdentity(text, size = 8) {
+  return sha256(text).toString().substring(0, size);
```
abompard, line 21: I still don't understand the reasoning behind this. React does not mandate the key attribute to have fixed width, or to have a limited character set. I don't see a case where the `text` value would not suffice.

PR #698, opened 2025-08-31
Implement the global search feature in the frontend

### LINKED ISSUE #689: Implement the global search feature in the frontend
Implement the global search feature in the frontend

Depends on ticket #665

gridhead:
Implement the global search feature in the frontend

Fixes #689


gridhead:
Active lookup works like the following

image

Ref. https://badges.gridhead.net/assembly

Press ENTER on active lookup to get this

image

Ref. https://badges.gridhead.net/discover/stop

As queries with insufficient information can be both wasteful and expensive, I have enforced an artificial restriction. The query length after trimming should be at least four characters before both the active lookup and search page works. 

Apart from that, I have gone ahead and consolidated a certain set of components based on their similar functionalities and simplified the URL path levelling for now. This should help with streamlining navigation for more complicated paths.

frontend/src/components/discover.jsx
```diff
@@ -0,0 +1,108 @@
...
+          {isLoading && (
+            <Dropdown.Item disabled className="small d-flex align-items-center">
+              <Spinner size="sm" className="me-2" />
+              Searching...
+            </Dropdown.Item>
+          )}
+
+          {error && (
+            <Dropdown.Item disabled className="small text-danger">
+              Search failed
```
laxathom, line 85: All those texts should be internationalized (e.g.  i18next)

frontend/src/routes/findpage.jsx
```diff
@@ -0,0 +1,111 @@
...
+    } else {
+      dispatch(hideLoad());
+    }
+  }, [isLoading, dispatch]);
+
+  if (findtext.length < 4) {
+    return <Mistaken />;
+  }
+
+  if (error) {
```
laxathom, line 31: Both findtext. and error can be factorized into one statement.

frontend/src/routes/findpage.jsx
```diff
@@ -0,0 +1,111 @@
...
+  const { data: dict, isLoading, error } = useRetrieveDiscoverQuery(findtext, { skip: false });
+
+  // Show or Hide LoadNote
+  useEffect(() => {
+    if (isLoading) {
+      dispatch(showLoad());
+    } else {
+      dispatch(hideLoad());
+    }
+  }, [isLoading, dispatch]);
```
laxathom, line 25: dispatch can be removed from dependency.

frontend/src/routes/findpage.jsx
```diff
@@ -0,0 +1,111 @@
...
+  if (findtext.length < 4) {
+    return <Mistaken />;
+  }
+
+  if (error) {
+    return <Mistaken />;
+  }
+
+  if (isLoading || !dict) {
+    return null;
```
laxathom, line 36: Is this a failsafe before the useEffect rendering?

frontend/src/routes/findpage.jsx
```diff
@@ -0,0 +1,111 @@
...
+          </Card.Body>
+        </Card>
+      </div>
+      <div className="col-12 col-lg-9">
+        <Card className="mb-2">
+          <Card.Body className="ps-0 pe-0 pt-2 pb-0">
+            <Card.Title className="mb-0 ps-2 dataelem" style={{ textTransform: "capitalize" }}>
+              Badges
+            </Card.Title>
+            <Card.Text className="mb-0 ps-2 small">Found {dict.badges.length} badge(s)</Card.Text>
```
laxathom, line 55: Same comment for the i18n feature

gridhead:
@laxathom i18n definitely needs to be factored in but is unfortunately out of the scope of this pull request. I will include your other suggestions though, so thanks a lot for those <3

PR #703, opened 2025-09-02
Create the UI for the leaderboards report page

### LINKED ISSUE #668: Create the UI for the leaderboards report page
Create the UI for the leaderboards report page

gridhead:
Create the UI for the leaderboards report page

Fixes #668

### DIFF
--- a/frontend/src/routes/rankings.jsx
+++ b/frontend/src/routes/rankings.jsx
@@ -1,0 +1,232 @@
+import { useEffect } from "react";
+import { Badge, Button, Card, Form, ListGroup } from "react-bootstrap";
+import { useDispatch, useSelector } from "react-redux";
+import { Link, useLocation, useNavigate, useParams } from "react-router";
+
+import VertItem from "../components/vertitem.jsx";
+import { useRetrieveRankingsQuery } from "../features/call.js";
+import { hideLoad, keepDate, showLoad } from "../features/part.js";
+import { generateIdentity, portraitProvider } from "../features/util.js";
+import Mistaken from "./mistaken.jsx";
+
+export default function Rankings() {
+  const dispatch = useDispatch();
+  const navigate = useNavigate();
+  const location = useLocation();
+  const thisdate = new Date();
+  const { y, m, d } = useParams();
+  const isweekly = location.pathname.endsWith("/week");
+  const pickdate = useSelector((state) => state.area.date);
+
+  const readDate = () => {
+    if (y && m && d) {
+      return `${y}-${m.toString().padStart(2, "0")}-${d.toString().padStart(2, "0")}`;
+    }
+    return pickdate || "";
+  };
+
+  const handleChange = (event) => {
+    const date = event.target.value;
+    dispatch(keepDate(date));
+
+    if (date) {
+      const [year, month, day] = date.split("-");
+      const weekSuffix = isweekly ? "/week" : "";
+      navigate(`/rankings/y/${year}/m/${parseInt(month)}/d/${parseInt(day)}${weekSuffix}`);
+    } else {
+      navigate("/rankings");
+    }
+  };
+
+  const params = {
+    ...(y && { y: parseInt(y) }),
+    ...(m && { m: parseInt(m) }),
+    ...(d && { d: parseInt(d) }),
+    ...(isweekly && { w: true }),
+  };
+
+  const { data: dict, isLoading, error } = useRetrieveRankingsQuery(params, { skip: false });
+
+  // Show or Hide LoadNote
+  useEffect(() => {
+    if (isLoading) {
+      dispatch(showLoad());
+    } else {
+      dispatch(hideLoad());
+    }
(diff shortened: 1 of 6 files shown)


gridhead:
The following URL routes are allowed.

The images did not load because I could not bother waiting for them to load up.

1. Daily - For eg. https://badges.gridhead.net/rankings/y/2025/m/8/d/4
   image

2. Weekly - For eg. https://badges.gridhead.net/rankings/y/2025/m/8/d/4/week
   image

3. Monthly - For eg. https://badges.gridhead.net/rankings/y/2025/m/8
   image

4. Yearly - For eg. https://badges.gridhead.net/rankings/y/2025
   image

5. All time - For eg. https://badges.gridhead.net/rankings
   image

6. Here is a convenient date picker utility without any additional library
   image

tahrir/views/report.py
```diff
@@ -135,61 +135,63 @@ def report_this_month():
...
-            # Yearly report
-            start = date(year, 1, 1)
-            stop = date(year, 12, 31)
+    begin = int(request.args.get('begin', 0))
+    limit = int(request.args.get('limit', 200))
+    if request.path.endswith('/week'):
+        # Weekly report
+        start = get_start_week(year, month, day)
+        stop = start + timedelta(days=6)
+        print("Start: ", start, " End: ", stop)
```
abompard, line 149: Leftover print statement

tahrir/views/report.py
```diff
@@ -135,61 +135,63 @@ def report_this_month():
...
+    data = [
+        {
+            "mail": user.avatar,
+            "nickname": user.nickname,
+            "badges": user_to_rank[user]["badges"],
+            "rank": {
+                "global": user.rank,
+                "period": user_to_rank[user]["rank"],
+            },
+        } for user in limited_users
```
abompard, line 195: This is an API change, but I don't think many people are using the badges API (un)fortunately.
It would be useful to also add the `user.id` to have a unique reference for the user. In React, it can be used for the `key` attribute [as recommended by the docs](https://react.dev/learn/rendering-lists#where-to-get-your-key).

frontend/src/routes/rankings.jsx
```diff
@@ -0,0 +1,208 @@
...
+      <div className="col-12 col-lg-9">
+        <Card className="mb-2">
+          <Card.Body className="ps-0 pe-0 pt-2 pb-0">
+            <Card.Title className="mb-0 ps-2 dataelem" style={{ textTransform: "capitalize" }}>
+              {showHead()}
+            </Card.Title>
+            <Card.Text className="mb-0 ps-2 small">Showing top performers</Card.Text>
+            <hr className="mt-2 mb-0" />
+            <ListGroup variant="flush">
+              {dict && dict.length > 0 ? (
```
abompard, line 208: What happens if `length` is zero? I think it won't produce any output already (but I may be wrong).

frontend/src/routes/rankings.jsx
```diff
@@ -0,0 +1,208 @@
...
+    if (conf === "d") link = `/rankings/y/${date.getFullYear()}/m/${date.getMonth() + 1}/d/${date.getDate()}`;
+    else if (conf === "w") link = `/rankings/y/${date.getFullYear()}/m/${date.getMonth() + 1}/d/${date.getDate()}/week`;
+    else if (conf === "m") link = `/rankings/y/${date.getFullYear()}/m/${date.getMonth() + 1}`;
+    else if (conf === "y") link = `/rankings/y/${date.getFullYear()}`;
+    else link = `/rankings`;
+    return link;
+  };
+
+  const scanDate = () => {
+    const today = new Date();
```
abompard: It looks like this function is called many times in the rendering. Could it be just called once and the result stored in a variable?
Also, what happens if the page is left open for more than a day and the user goes back to it? The date will not be refreshed, right?

abompard (changes requested):

tahrir/views/report.py
gridhead: Dictionaries, when they are returned as an HTTP response, have a likelihood of losing their initial sorting to an order that makes sense to the end receiving them. In this case, if we were to return a dictionary (using this term loosely for a map between the nickname and the information), we would lose the ordering with respect to the rankings. We can always iterate through the list for the nicknames and use them as `key`s for React rendering so I reckon that these stay put.

frontend/src/routes/rankings.jsx
gridhead: This, if I understood your question correction https://github.com/fedora-infra/tahrir/pull/703/files#diff-e25eb7f4400fd68a3a86343afb86341009b329b8f19a3a51e3facb83f0b42087R200. 

image

gridhead:
@abompard, I made the suggested changes. Feel free to give it another look.

gridhead:
Taking the liberty of merging this without approval now as @abompard would be available more in the coming weeks.

PR #704, opened 2025-09-03
Devise a magical script for computing rarities of the collectable badges

### LINKED ISSUE #700: Devise a magical script for computing rarities of the collectable badges
Devise a magical script for computing rarities of the collectable badges

gridhead:
Devise a magical script for computing rarities of the collectable badges

Fixes #700


gridhead:
The file looks like the following

https://github.com/fedora-infra/tahrir/blob/bd857c0d57e55fd37bdb137dac8c0bad8d4ffba3/tahrir/static/rarities.json

And the magic looks like the following

c3

I expect this script to be run once per Fedora Linux release cycle to generate the rarities of the badges, and this should keep things interesting around the Fedora Project contributor community.

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,57 @@
...
+
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+
+from tahrir_api.dbapi import TahrirDatabase
+from tqdm import tqdm
+
+
+def generate_rarities():
+    database_link = os.environ.get("SQLALCHEMY_DATABASE_URI", "postgresql://tahrir:tahrir@192.168.0.212:5432/tahrir")
```
abompard: Tahrir has a configuration file, you could make this script use it.

That said, this script would probably be better located in the `tahrir-api` package, since it's unrelated to the web interface and only works with the database.

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,57 @@
...
+        accolist = database.get_all_badges().all()
+        userlist = database.get_all_persons().all()
+        accodict = {}
+
+        pbar = tqdm(enumerate(accolist, 1), total=len(accolist), colour="#008080", leave=False, bar_format='{desc} {percentage:4.2f}%|{bar}| {n_fmt}/{total_fmt} [{elapsed}<{remaining}, {rate_fmt}]')
+        for iter, unit in pbar:
+            pbar.set_description(f"Processing {unit.name[:20]:<20}")
+            accodict[unit.id] = {
+                "poll": len(database.get_assertions_by_badge(unit.id)),
+                "rate": (len(database.get_assertions_by_badge(unit.id)) / len(userlist) * 100)
```
abompard: You're making the `get_assertions_by_badges()` call twice here.

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,57 @@
...
+            }
+
+        accodict = dict(sorted(accodict.items(), key=lambda item: item[1]["poll"]))
```
abompard: If you convert to a dict, you loose the sorting. You may want to use an [OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict) instead.

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,57 @@
...
+        rarities = {"X": [], "S": [], "A": [], "B": [], "C": [], "D": []}
+        size = len(accodict.keys()) // len(rarities.keys())
+        left = len(accodict.keys()) % len(rarities.keys())
+        jump = 0
+        for iter, rare in enumerate(rarities.keys()):
+            stop = jump + size + (1 if iter < left else 0)
+            for iden, data in list(accodict.items())[jump: stop]:
+                data["rare"] = rare
+                rarities[rare].append(iden)
+            jump = stop
```
abompard: This block will need detailed comments to explain what's going on.

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,57 @@
...
+
+        jsondata = {
+            "rarity": rarities,
+            "badges": accodict,
+        }
+        with open("rarities.json", "w") as json_file:
```
abompard: Why not store this in the DB? You could put the `rare` value in the `badges` table, no?

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,57 @@
...
+        print("Rarities saved to 'rarities.json' - Commit the generated file")
+    except Exception as expt:
+        print(f"Rarities generation failed: {expt}")
+        sys.exit(1)
+    finally:
+        sessobjc.close()
+
+
+if __name__ == "__main__":
+    generate_rarities()
```
abompard: You don't need this if you've added the script to `tools.poetry.scripts`.

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,57 @@
...
+            jump = stop
+
+        jsondata = {
+            "rarity": rarities,
+            "badges": accodict,
+        }
+        with open("rarities.json", "w") as json_file:
+            json.dump(jsondata, json_file, indent=4, sort_keys=False)
+        print("Rarities saved to 'rarities.json' - Commit the generated file")
+    except Exception as expt:
```
abompard, line 80: Instead of a giant try/except, I usually recommend putting the content inside a separate function (that would here take the session as argument) and having the try/except around the function call. I find it easier to match the `except` to its `try`.

Also, if you prefer I think you can do `with sessobjc() as session:` to have a contextual block that will automatically close the session and raise the exception if there is any.

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,57 @@
...
+
+    try:
+        sessobjc = sessionmaker(bind=create_engine(database_link))()
+        database = TahrirDatabase(session=sessobjc, autocommit=False)
+        accolist = database.get_all_badges().all()
+        userlist = database.get_all_persons().all()
+        accodict = {}
+
+        pbar = tqdm(enumerate(accolist, 1), total=len(accolist), colour="#008080", leave=False, bar_format='{desc} {percentage:4.2f}%|{bar}| {n_fmt}/{total_fmt} [{elapsed}<{remaining}, {rate_fmt}]')
+        for iter, unit in pbar:
```
abompard: It doesn't look like you're actually using the `iter` variable in this block.

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,57 @@
...
+            pbar.set_description(f"Processing {unit.name[:20]:<20}")
+            accodict[unit.id] = {
+                "poll": len(database.get_assertions_by_badge(unit.id)),
+                "rate": (len(database.get_assertions_by_badge(unit.id)) / len(userlist) * 100)
+            }
+
+        accodict = dict(sorted(accodict.items(), key=lambda item: item[1]["poll"]))
+        rarities = {"X": [], "S": [], "A": [], "B": [], "C": [], "D": []}
```
abompard: The list of rarities names could be a top-level variable, so that we can easily change it if we want to. You would then build the same structure with `rarities = {name: [] for name in RARITIES}`

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def generate_rarities():
+    database_link = os.environ.get("SQLALCHEMY_DATABASE_URI", "postgresql://tahrir:tahrir@192.168.0.212:5432/tahrir")
+
+    try:
+        sessobjc = sessionmaker(bind=create_engine(database_link))()
+        database = TahrirDatabase(session=sessobjc, autocommit=False)
+        accolist = database.get_all_badges().all()
+        userlist = database.get_all_persons().all()
```
abompard: You're apparently only using this variable to get its `len()`, so instead of loading the entire result in memory you could do `database.get_all_persons().count()`.

abompard (changes requested):

tahrir/utils/rarities.py
gridhead: It is used here https://github.com/fedora-infra/tahrir/pull/704/files#diff-f7ff9cfb5c0c5552125e1f903f419ab2a4a141e6783f111ffba9e02a7ceddab7R23.

abompard: Not sure I undersand. You're iterating on `enumerate()`, which gives you the `(index, item)` instead of just `item`, but you're apparently not using the `index` value in the block below. So you could just iterate on `accolist` directly and you would have `for unit in pbar:`.

gridhead:
Ok, I am made some changes.

Here's how the script can be invoked, for example.

```
(venv) $ FLASK_CONFIG=/home/tinyhost/Projects/tahrir/devel/ansible/roles/dev/files/tahrir.cfg rarities
```

Regarding https://github.com/fedora-infra/tahrir/pull/704#discussion_r2321462933 (and the second half of https://github.com/fedora-infra/tahrir/pull/704#discussion_r2321424835), I have opened a ticket https://github.com/fedora-infra/tahrir-api/issues/193 to track this feature and will probably retire from here once it is implemented there.

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,85 @@
...
+    For e.g., at the time of writing this code, there is a total of 628 badges and I created
+    6 rarity levels, starting from X (most rare) to D (least rare). The dictionary of 
+    badges is sorted with respect to the assertion count and then, equal segments of sizes
+    equalling ~ (628 // 6) are created.
+
+    As 628 does not divide neatly by 6, the size of each rarity becomes 104 (i.e. 628 // 6)
+    before the remainder 4 (i.e. 628 % 6) is also spread through all the rarities from X 
+    onwards until we run out of it. This ends up in a division where X, S, A and B rarities
+    have 105 badges and C and D rarities have 104 badges.
+    """
```
abompard, line 53: Thanks!! Much clearer to me :-)

Having rarities of equal sizes is an interesting approach, there would also be an argument for having less of the rarer, for example X has 5%, S has the next 10%, A has the next 15%, B has the next 20%, C has the next 25% and D has the rest (about 25% too). Or something like this. But equal size has merits too, you decide :-)

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,85 @@
...
+    accodict = OrderedDict(sorted(accodict.items(), key=lambda item: item[1]["poll"]))
+    rarities = {name: [] for name in RARITIES}
+    size = len(accodict.keys()) // len(rarities.keys())
+    left = len(accodict.keys()) % len(rarities.keys())
```
abompard: By construction, `rarities.keys() == RARITIES` :-)

```suggestion
    size = len(accodict.keys()) // len(RARITIES)
    left = len(accodict.keys()) % len(RARITIES)
```

tahrir/utils/rarities.py
```diff
@@ -0,0 +1,85 @@
...
+            rarities[rare].append(iden)
+        jump = stop
+
+    jsondata = {
+        "rarity": rarities,
+        "badges": accodict,
+    }
+    with open("rarities.json", "w") as json_file:
+        json.dump(jsondata, json_file, indent=4, sort_keys=False)
+    print("Rarities saved to 'rarities.json' - Commit the generated file")
```
abompard: OK let's wait for the database column, I don't think we should commit in the source code the result of this computation, as it is very fedora-specific, temporary, and depending on data that is not inside the source code (we want a separation of source code and data).

abompard (changes requested):

tahrir/utils/rarities.py
gridhead: It's not like people apart from Fedora Project use Tahrir but I agree that the source code and data assets should be separated.

gridhead:
@abompard, I made the suggested changes. Feel free to give it another look.

gridhead:
Taking the liberty of merging this without approval now as @abompard would be available more in the coming weeks.

PR #744, opened 2025-10-07
Add CRUD REST API endpoints for PERSONS relation

### LINKED ISSUE #694: Implement the necessary endpoints for performing CURD operations on the `PERSONS` relation
Implement the necessary endpoints for performing CURD operations on the `PERSONS` relation

Related to #680

sdglitched:
This PR adds REST API endpoints for user (`PERSONS`) relation.

Key Changes:
 - Added new endpoint for fetching specific user details including badges, invitations, rank, and percentile
 - Added new endpoint for generating the diff between two users (the one who want to compare has to be logged in)
 - Added new endpoint for adding user in the database. I'm very sceptical about this one. If we are using OIDC for the user account creation, shouldn't only user be able to create the account? Why do admin be able to do it?
- Added new endpoint for deactivating or deactivating users account

Endpoints added:
- GET /api/users/ - Fetch specific user details including badges, invitations, rank, and percentile
- GET /api/users/diff/ - Compare badges between two users, showing unique and shared badges (requires login)
- POST /api/admin/users - Adds new user the the database 
- PUT /api/users - Deactivates or reactivates users account (requires login)

Note: Full CRUD operations require new database functions in [tahrir-api](https://github.com/fedora-infra/tahrir-api/issues/200) first:
 - Delete user - for DELETE operations
 https://github.com/fedora-infra/tahrir-api/blob/6c36987df0593937f4c061d1424c0dd426e43891/tahrir_api/dbapi.py#L547C1-L560C21
 Delete is present but CASCADE delete is not enable in the database
 - Update user - for UPDATE operations

Thank you for reviewing this PR and let me know if any modifications are needed.

### DIFF
--- a/tahrir/endpoints/admin/users.py
+++ b/tahrir/endpoints/admin/users.py
@@ -1,0 +1,33 @@
+from flask import abort, g, jsonify, request
+
+from ...app import csrf, oidc
+from ...utils.user import require_admin
+from . import blueprint as bp
+
+
+@bp.route("/api/admin/users", methods=["POST"])
+@csrf.exempt
+@oidc.require_login
+@require_admin
+def add_user():
+    """Endpoint to add a new user"""
+
+    data = request.get_json()
+    if not data:
+        return abort(400, "No details provided")
+
+    if not data.get("email"):
+        return abort(400, "No email id is provided")
+
+    if g.tahrirdb.person_exists(email=data.get("email")):
+        abort(409, f"Person with email {data.get('email')!r} already exists.")
+
+    g.tahrirdb.add_person(
+        email=data.get("email"),
+        nickname=data.get("nickname"),
+        website=data.get("website"),
+        bio=data.get("bio"),
+        avatar=data.get("avatar"),
+    )
+
+    return jsonify({"message": f"User {data.get('email')!r} added successfully"}), 201
--- a/tahrir/endpoints/users.py
+++ b/tahrir/endpoints/users.py
@@ -1,7 +1,8 @@
 import sqlalchemy as sa
 import tahrir_api.model as m
-from flask import g, jsonify, request
+from flask import abort, g, jsonify, request
 
+from ..utils.user import get_person, get_user_badge_info, require_login
 from . import blueprint as bp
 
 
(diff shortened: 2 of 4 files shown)


sdglitched:
@abompard and @gridhead, thank you for reviewing.

tahrir/endpoints/users.py
```diff
@@ -44,3 +47,142 @@ def search_users_by_string(search_string: str):
...
+            "badge_id": inv.badge_id,
+            "expires_on": inv.expires_on.isoformat(),
+            "created_on": inv.created_on.isoformat(),
+        }
+        for inv in g.tahrirdb.get_invitations(user.id)
+        if inv.expires_on > datetime.now()
+    ]
+
+    # Get badge info using utility function
+    badges_info = get_user_badge_info(user)
```
gridhead, line 63: Can the `badges_by_tag` be a list of indices that point to the elements inside of the `badges` list? That should help de-duplicate the same set of information passed down twice?

gridhead, line 63: Also, as we have the `badges` list that uncategorically lists all the badges that a user has, I do not think we need that `uncategorized_badges` list anymore.

tahrir/endpoints/users.py
```diff
@@ -44,3 +47,142 @@ def search_users_by_string(search_string: str):
...
+    invitations = [
+        {
+            "id": inv.id,
+            "badge_id": inv.badge_id,
+            "expires_on": inv.expires_on.isoformat(),
+            "created_on": inv.created_on.isoformat(),
+        }
+        for inv in g.tahrirdb.get_invitations(user.id)
+        if inv.expires_on > datetime.now()
+    ]
```
gridhead: I was thinking as to where we could list this thing. Maybe we could show it in the users settings page which they can access when they are logged in? As such, please limit this information behind login (as in, if they are logged in, only then the invitations would be shown to them). After all, there is no use of me being able to see somebody else's invitations, you know.

tahrir/endpoints/users.py
```diff
@@ -44,3 +47,142 @@ def search_users_by_string(search_string: str):
...
+    # Get badge info using utility function
+    badges_info = get_user_badge_info(user)
+
+    return jsonify(
+        {
+            "user": {
+                "id": user.id,
+                "nickname": user.nickname,
+                "email": user.email,
+                "avatar": user.avatar,
```
gridhead:
```suggestion
                "mail": user.avatar,
```

As the email can be constructed using the `username`, there is no point in passing it. We only need the `avatar` here in order to be able to retrieve their portraits from Libravatar.

Maybe in the future, instead of passing down mails like these, we should hash them and pass them down to ensure that we do not serve email addresses to spammers on a silver platter.

tahrir/endpoints/users.py
```diff
@@ -44,3 +47,142 @@ def search_users_by_string(search_string: str):
...
+                "created_on": user.created_on.isoformat() if user.created_on else None,
```
gridhead: Please send the Epoch time here. All the computations would be taken care of by the frontend's JavaScript to convert the time in accordance with the browser's timezone.

tahrir/endpoints/users.py
```diff
@@ -44,3 +47,142 @@ def search_users_by_string(search_string: str):
...
+                "opt_out": user.opt_out,
```
gridhead, line 71: I do not think there is a need to separately list this. If the user can be found, that essentially means that they have an account that exists and have not opted out.

tahrir/endpoints/users.py
```diff
@@ -44,3 +47,142 @@ def search_users_by_string(search_string: str):
...
+        if inv.expires_on > datetime.now()
+    ]
+
+    # Get badge info using utility function
+    badges_info = get_user_badge_info(user)
+
+    return jsonify(
+        {
+            "user": {
+                "id": user.id,
```
gridhead: What's the difference between this and the `user_count` that is shown inside of the `badges_info` section? I would suggest that both of these have no business being shown as these things are for internal database reference only.

tahrir/endpoints/users.py
```diff
@@ -44,3 +47,142 @@ def search_users_by_string(search_string: str):
...
+def get_user_by_id(user_id: str):
+    """Endpoint to fetch the user based on matching id."""
+
+    user = get_person(user_id)
+
+    if not user:
+        abort(404, f"No such user {user_id!r}")
+
+    if user.opt_out and user.email != g.oidc_user.email:
+        abort(404, f"User {user_id!r} has opted out.")
```
gridhead, line 60: @abompard you seem to be leaving some positive impact on @sdglitched here. I see more small blocks with early fails instead of large blocks with cascading conditions.

tahrir/endpoints/users.py
```diff
@@ -44,3 +47,142 @@ def search_users_by_string(search_string: str):
...
+def update_user_by_id():
+    """Endpoint to update user account settings."""
+
+    data = request.get_json()
+    if not data.get("opt_out"):
+        abort(400, "No data provided")
+
+    g.oidc_user.person.opt_out = data.get("opt_out")
+
+    return jsonify({"message": "User updated successfully"})
```
gridhead, line 90: I wonder if we should return a `202` here instead of `200`. Please note that both of them are usually correct and it is just a matter of precise responses.

tahrir/endpoints/users.py
```diff
@@ -44,3 +47,142 @@ def search_users_by_string(search_string: str):
...
+
+
+@bp.route("/api/users/diff/<string:id_a>/<string:id_b>", methods=["GET"])
+@require_login
```
gridhead, line 94: I am glad that this is behind a login. It should not be people's business to check the diffs between Person A and Person B when they aren't either of these persons.

gridhead, line 94: But... how do we confirm if the person that is logged in is indeed either of Person A or Person B? Diffs are usually made directionally so their orders also matter.

gridhead, line 94: While you are at it, could you please attach a response in diffs? As I have not yet made client-side JavaScript changes, this would help me in reviewing.

tahrir/endpoints/admin/users.py
```diff
@@ -0,0 +1,33 @@
...
+    if g.tahrirdb.person_exists(email=data.get("email")):
+        abort(409, f"Person with email {data.get('email')!r} already exists.")
+
+    g.tahrirdb.add_person(
+        email=data.get("email"),
+        nickname=data.get("nickname"),
+        website=data.get("website"),
+        bio=data.get("bio"),
+        avatar=data.get("avatar"),
+    )
```
gridhead, line 31: I will keep myself from reviewing this part until this gets merged and I make the necessary client-side changes to work with this administration endpoint.

tahrir/utils/user.py
```diff
@@ -152,3 +154,55 @@ def decorated(*args, **kwargs):
...
+    total_badges = g.tahrirdb.get_all_badges().count()
+
+    # Get percentage of badges earned
+    percent_earned = (float(len(badges)) / float(total_badges)) * 100 if total_badges > 0 else 0
+
+    # Get rank
+    rank = user.rank or 0
+
+    # Get total number of users
+    user_count = g.tahrirdb.get_all_persons().count()
```
gridhead, line 177: See https://github.com/fedora-infra/tahrir/pull/744/files#r2425456541. 

I could use some more context here.

tahrir/utils/user.py
```diff
@@ -152,3 +154,55 @@ def decorated(*args, **kwargs):
...
+    return {
+        "badges": serialized_badges,
+        "total_badges": total_badges,
+        "percent_earned": round(percent_earned, 2),
+        "rank": rank,
+        "user_count": user_count,
+        "percentile": percentile,
+        "badges_by_tag": serialized_badges_by_tag,
+        "uncategorized_badges": serialized_uncategorized_badges,
+    }
```
gridhead, line 205: Please look into the suggestions that I have made against the `/api/users/<username>` endpoint as a bunch of those actually end up applying here.

gridhead (changes requested):
Please make the following changes. Thanks.

tahrir/endpoints/users.py
sdglitched: That does makes sense, their is no need for someone to view the invitations of someone else. If we want to have it in the user's settings page, then it would be better to have it as a different endpoint in `Invitation` CRUD.

I'm removing it for now, we can revisit it once the endpoint is needed for it.

tahrir/endpoints/users.py
sdglitched: I believe having this is necessary. Let's say I opted out but I want back in, then the logged in version need to show me some kind of toggle/button which will allow me to re-activate my account. And for getting the info from backend, the field is essential.

tahrir/endpoints/users.py
sdglitched: I would have done 202 had we been using `fastapi` 😛

tahrir/endpoints/users.py
sdglitched: This is the best I can give.. admin login is of `aaronhale` 
image

sdglitched:
@gridhead, I made the changes which I see fit, thank you for reviewing it again. Also, I've added a small change in `/endpoints/assertions.py` which is unrelated to this, but is an essential change.

gridhead (approved):
LGTM. Thanks for your contribution.

PR #827, opened 2026-03-22
fix: include last day of month in monthly leaderboard

### LINKED ISSUE #320: Badges Monthly Leaderboard misses the last day
Badges acquired on last day of the month do not show up on Monthly Leaderboard resulting in wrong ranking.
Eg :
Monthly LeaderBoard Sept 2014 : https://badges.fedoraproject.org/report/2014/9 
It shows @robyduck has earned 16 badges during the period
But if you check out @robyduck 's badges here : https://badges.fedoraproject.org/user/robyduck?history_limit=115
@robyduck has earned 16 badges before 30 Sept + 7 badges on 30 Sept , effectively making @robyduck #3 on Monthly LeaderBoard

Same goes for @bee2502 for Oct 2015 : https://badges.fedoraproject.org/user/bee2502
Badges earned during Oct = 21 = 19 + 2(on 31st)
Monthly Leaderboard still shows 19 badges : https://badges.fedoraproject.org/report/2015/10

Marvinrose:
## Bug fix: Monthly leaderboard missing last day of month
Fixes #320 

### Root cause: 
In `report_year_month()`, `stop.replace(day=1)` was called but its return value was never saved back to `stop`. Python's `date.replace()` returns a new object rather than modifying in place, so the result was silently discarded and `stop` remained incorrectly set.

### Fix: 
Replaced the broken calculation with the correct approach, consistent with how `home()` already calculates the last day of the month.
The old Pyramid-based `views.py` used `date(year, month + 1, 1) - timedelta(days=1)` which confirms the original intent. However, that approach had a latent bug for December (month + 1 = 13 is invalid). This fix handles that edge case correctly.
-  Added tests

### DIFF
--- a/tahrir/views/report.py
+++ b/tahrir/views/report.py
@@ -58,9 +58,10 @@ def report_year_month(year, month):
 
     start = date(year, month, 1)
     # get the last day of the month
-    stop = start + timedelta(days=32)
-    stop.replace(day=1)
-    stop = stop - timedelta(days=1)
+    if month == 12:
+        stop = date(year + 1, 1, 1) - timedelta(days=1)
+    else:
+        stop = date(year, month + 1, 1) - timedelta(days=1)
 
     user_to_rank = g.tahrirdb.make_leaderboard(
         start=start,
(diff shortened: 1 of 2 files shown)


Marvinrose:
A new PR has been opened targeting the swatantrya branch instead: #874

PR #834, opened 2026-03-23
feat: replace badge builder with WTForms and Jinja2 YAML generation

### LINKED ISSUE #520: Replace the badge builder form
The Badge Builder form should use WTForms to make it easier to maintain and more standard (such as when the posted fields are not reused on the resulting page).
It should have select boxes where options are limited, such as with the Condition.
It should do the YAML generation in the Jinja2 template instead of python string templating to have better conditionals.

Marvinrose:
## feat: Replace badge builder form with WTForms and Jinja2 YAML generation

Closes #520 

### Changes
- Added `BadgeBuilderForm` in `tahrir/forms.py` using Flask-WTF
- `Condition` field is now a `SelectField` dropdown instead of free text input
- YAML generation moved from Python string templating in generate_badge_yaml()` to Jinja2 template conditionals
- Removed `generate_badge_yaml()` from `tahrir/utils/badge.py`
- Form fields are no longer reused on the result page
- CSRF protection handled properly via `form.hidden_tag()`
- Removed unused `builder.js` script reference

### DIFF
--- a/tahrir/forms.py
+++ b/tahrir/forms.py
@@ -1,0 +1,43 @@
+from flask_wtf import FlaskForm
+from wtforms import SelectField, StringField, TextAreaField
+from wtforms.validators import Optional
+
+
+DEFAULT_PREVIOUS = """\
+filter:
+  topics:
+    - message.topic
+  users:
+    - recipient
+  rows_per_page: 0
+operation: count"""
+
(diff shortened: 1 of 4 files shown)


Marvinrose:
A new PR has been opened targeting the swatantrya branch instead: #876

PR #836, opened 2026-03-23
Fix/badgr export 414 chunked urls

### LINKED ISSUE #486: Export to Badgr fails with error 414 Request-URI Too Large
The URL that Tahrir builds to export the badges can become too large, and Badgr refuses it.
Maybe we can find a way to send the badges in smaller chunks. Javascript will probably be involved in this fix.

Marvinrose:
Fixes #486 

When exporting many badges to Badgr, the generated URL could exceed the server's maximum URI length, resulting in a 414 error.

Fixed by splitting badge URLs into batches of 10 before calling OpenBadges.issue(). For the modal version, batches are processed sequentially using callbacks. For the no-modal version, batches are sent in a loop.

### DIFF
--- a/tahrir/templates/_base.html
+++ b/tahrir/templates/_base.html
@@ -37,11 +37,28 @@
         }
         function claim_badges() {
           var urls = badge_urls();
-          {% if config['TAHRIR_OPENBADGES_MODAL'] %}
-          var lolback = function(errors, successes) {};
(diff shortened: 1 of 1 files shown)


Marvinrose:
A new PR has been opened targeting the swatantrya branch instead: #877

PR #841, opened 2026-03-23
Fix add_tag view to not accept request as parameter

### LINKED ISSUE #840: `add_tag` view crashes with TypeError, Flask view incorrectly accepts `request` parameter
## Description

The `add_tag` view function in `tahrir/views/admin.py` (line 227) is defined as:

```python
def add_tag(request):
```

Flask view functions do not receive `request` as a positional argument — it should be accessed via `from flask import request` (which is already imported at the top of the file). This causes the endpoint to **always crash with a TypeError** because Flask passes zero positional arguments to the view function.

## Steps to Reproduce

1. Log in as an admin
2. Navigate to a badge page
3. Attempt to add a tag via the form (POST to `/add_tag`)
4. The endpoint crashes with: `TypeError: add_tag() missing 1 required positional argument: 'request'`

## Expected Behavior

The tag should be added to the badge and the user should be redirected back to the badge page.

System625:
Fixes #840 
The add_tag view was defined as def add_tag(request): but Flask views don't receive request as a positional argument, causing a TypeError on every call. Also adds tests and fixes dogpile cache test isolation in conftest.

### DIFF
--- a/tahrir/views/admin.py
+++ b/tahrir/views/admin.py
@@ -224,7 +224,7 @@ def award_from_csv():
 @bp.route("/add_tag", methods=["POST"])
 @oidc.require_login
 @require_admin
-def add_tag(request):
+def add_tag():
     badge_id = request.form.get("badge_id")
     badge = g.tahrirdb.get_badge(badge_id)
     if not badge:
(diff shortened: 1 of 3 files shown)


System625:
Closing as instructed, will retarget from swatantrya branch

PR #826, opened 2026-03-21
Return hashed email address instead of actual email address for  photographs

System625:
Resolves #749

PR #844, opened 2026-03-23
Return hashed email instead of raw email in JSON responses

### LINKED ISSUE #749: Return hashed email address instead of actual email address for photographs
Return hashed email address instead of actual email address for photographs

System625:
Fixes #749
Second version of #826 targeting swantantrya branch
Assisted by Claude(Anthropic)

### DIFF
--- a/frontend/src/features/util.js
+++ b/frontend/src/features/util.js
@@ -1,4 +1,3 @@
-import md5 from "crypto-js/md5";
 import sha256 from "crypto-js/sha256";
 
 export function formatTime(epochTime) {
@@ -14,7 +13,7 @@ export function formatTime(epochTime) {
 }
 
 export function portraitProvider(mail, size = 40) {
-  return `https://seccdn.libravatar.org/avatar/${md5(mail.trim().toLowerCase()).toString()}?s=${size}&d=retro`;
+  return `https://seccdn.libravatar.org/avatar/${mail}?s=${size}&d=retro`;
 }
 
 export function generateIdentity(text, size = 8) {
--- a/frontend/src/routes/homepage.jsx
+++ b/frontend/src/routes/homepage.jsx
@@ -194,7 +194,7 @@ export default function Homepage() {
                       )}
                     </>
                   }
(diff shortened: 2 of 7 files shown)


gridhead (changes requested):
Please review the [AI-Assisted Contributions Policy](https://docs.fedoraproject.org/en-US/council/policy/ai-contribution-policy/) and make necessary changes to disclose the use of AI tools.

gridhead (changes requested):
Also, @System625, please resolve the merge conflicts while you are at it.

gridhead (changes requested):
Are you certain that the hashed email is used by the frontend in the way it is supposed to be?

System625:
Hello @gridhead , sorry about that. When I initially reviewed the fix, I verified the backend was storing and returning the hashed email correctly, but I hadn't fully traced how the frontend consumed it. So after your comment, I went deeper with the help of Claude and I found that portraitProvider in util.js was applying md5() on the incoming mail value, which meant when it received a SHA256 hash instead of a raw email, it was computing md5(sha256_hash) producing a completely wrong Libravatar URL. I've now updated portraitProvider to detect pre-hashed values and use them directly. Please kindly review again and let me know if I missed something else

gridhead (commented):
@System625, I elected to retain this pull request and introduce corrective changes here for this to be a learning opportunity for you. While you applied your fixes to the wrong layers of the code (i.e. hashing when the person's entry is created), you also needlessly introduced double-hash checks everywhere in the code as if one part of the codebase does not trust the other part of the codebase enough. This is a hallmark behaviour of someone who has not spent enough time understanding the functionality of the codebase, and this is your opportunity to change that.

Your changes also did not include migration scripts, so had we gone on that route, your changes would only have helped the newly created users, but worse still, broken operations for the existing users. The `portraitProvider` on the frontend would have still computed MD5 hashes based on the SHA256 hashes sent from the server side, thus causing quite the messy affair. But since the endpoints happened to be the correct place to resolve this problem, I decided to undo your mistaken commits and add my corrective changes in their place, instead of rejecting your changes straightaway.

PR #858, opened 2026-03-25
fix: allow cache reconfiguration between tests

### LINKED ISSUE #857: `RegionAlreadyConfigured` error prevents running full test suite
## Problem
Running the full test suite with `pytest tests/` causes all tests using the `app` fixture to fail after the first test with:
```
dogpile.cache.exception.RegionAlreadyConfigured: This region is already configured with backend: NullBackend. Specify replace_existing_backend=True to replace.
```

## Steps to reproduce
```bash
pytest tests/ -v
```

## Root cause
`dogpile.cache` uses a module-level singleton. Each test creates a fresh app instance which calls `cache.configure()`, but the second call raises `RegionAlreadyConfigured` because the cache was already configured by the first test.

## Affected tests
`test_dbadmin_badge` and any other test using the `app` fixture.

## Expected behavior
All tests should pass when running `pytest tests/`

## Actual behavior
Only the first test using the `app` fixture passes. All subsequent ones fail at setup with `RegionAlreadyConfigured`.

## Proposed fix
Add `replace_existing_backend=True` to the `CACHE` config in `conftest.py`, this is the flag `dogpile.cache` itself provides for exactly this scenario

Marvinrose:
## fix: Allow cache reconfiguration between tests

Fixes #857 

### Problem
Running `pytest tests/` caused all tests using the `app` fixture to fail after the first test with:
```
dogpile.cache.exception.RegionAlreadyConfigured
```
The `dogpile.cache` region is a module-level singleton. Each test creates a fresh app instance which tries to reconfigure the cache, but the second call raises an error.

### Fix
Added `replace_existing_backend=True` to the `CACHE` config in `conftest.py`. This is the flag dogpile.cache itself provides for 
exactly this use case.

### Result
Before: `1 passed, 5 errors` (when running multiple tests)  
After: `11 passed, 0 failures`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,6 +22,11 @@ def app_config(tmpdir):
         TAHRIR_ADMIN_GROUPS=["admin"],
         OIDC_ENABLED=False,
         OIDC_TESTING_PROFILE={"nickname": "test-user", "groups": ["admin"]},
+        CACHE={
+            "backend": "dogpile.cache.null",
+            "expiration_time": 10000,
+            "replace_existing_backend": True,  # fix: allow reconfiguring cache between tests
+        },
     )
 
 


Marvinrose:
Hi @gridhead 

Kindly review..

Marvinrose:
Closing this PR as the fix (replace_existing_backend=True) is already present in the swatantrya branch's conftest.py. The issue has effectively been resolved in the revamped codebase.

PR #863, opened 2026-03-26
fix: allow tooltip/popover text selection by staying open during hover

### LINKED ISSUE #771: Implement text copying ability in the `tooltips`
Currently the `tooltips` vanishes once you move away from a certain hover-able object and coping the text from it is not possible.

Image

Daniel-1600:
Allow text selection in badge popovers

Badge info popovers were disappearing when hovering away, making it impossible to copy the text. Now they stay open while you interact with them, so you can select and copy the information.

closes #771

description

https://github.com/user-attachments/assets/139d1bdd-a30c-484f-ae86-5a94793c7c77


gridhead (approved):
Looks good to me. 🥇

PR #869, opened 2026-03-27
Sort badge holders names alphabetically

### LINKED ISSUE #382: Sort "Badge Holders" names alphabetically
Badge pages have a list of users who were awarded the badge. As someone who awards badges, this list is hard to parse for a previous awardee since it's not ordered. Ideally the list should be alphanumerically sorted so I can locate who's already been awarded the badge.

olamidepeterojo:
Fixes #382

### DIFF
--- a/tahrir/endpoints/assertions.py
+++ b/tahrir/endpoints/assertions.py
@@ -40,7 +40,7 @@ def get_assertions_by_badge(badge_id: str):
     if not assertions:
         return jsonify([])
 
-    assertions = sorted(assertions, key=lambda assertion: assertion.issued_on)
+    assertions = g.tahrirdb.get_assertions_by_badge(badge_id)
 
     # This is a very unoptimised implementation for achieving pagination.
     # The implementation should have been there in the upstream `tahrir-api` at database level.


tahrir/views/badge.py
```diff
@@ -49,8 +49,9 @@ def badge(badge_id):
         percent_earned = float(times_awarded) / float(len(g.tahrirdb.get_all_persons().all()))
 
         # This is a list of assertions for this badge.
-        badge_assertions = (
-            g.tahrirdb.get_all_assertions().filter(m.Assertion.badge_id == badge.id).all()
+        badge_assertions = sorted(
+            g.tahrirdb.get_assertions_by_badge(badge.id),
+            key=lambda a: (a.person.nickname or "").lower(),
         )
```
gridhead: Would it not make sense to have this sorting done inside of `tahrir-api` instead of over here? There is a certain set of badges with quite a lot of recipients, and as such, if the sorting is not done at a lower level, this could cause the application to slow down, as we are still making use of a synchronous framework.

tahrir/views/badge.py
gridhead: While I understand what you are going for over here, I still struggle to see the value of the introduced changes when they have been made in the legacy Jinja-based templating frontend and not on the modern ReactJS-based progressive frontend. We want folks to use your codebase as much as you do, after all.

gridhead (changes requested):
Please look into the suggested changes.

tahrir/views/badge.py
olamidepeterojo: yes it does

tahrir/views/badge.py
olamidepeterojo: My bad, the first changes were because I created my branch of develop instead of swatantrya

olamidepeterojo:
Hello @gridhead  @sdglitched , this fix comes from https://github.com/fedora-infra/tahrir-api/pull/270, which I proposed because @gridhead suggested I implement the fix at the DB level.

While testing, I found that the ordering from the DB was being overridden in `tahrir/endpoints/assertions.py`, where assertions were re-sorted by issued_on ` assertions = sorted(assertions, key=lambda assertion: assertion.issued_on)`. So I removed this sorting so that the sorting comes from the DB and is not altered.


Below is an image showing badge holders sorted

Screenshot from 2026-04-05 13-29-20

gridhead (changes requested):
See https://github.com/fedora-infra/tahrir/issues/382#issuecomment-4191693392.

Closing this as unplanned.

PR #877, opened 2026-03-28
fix: Badgr export 414 chunked urls

### LINKED ISSUE #486: Export to Badgr fails with error 414 Request-URI Too Large

Marvinrose:
Fixes #486 

When exporting many badges to Badgr, the generated URL could exceed the server's maximum URI length, resulting in a 414 error.

Fixed by splitting badge URLs into batches of 10 before calling OpenBadges.issue(). For the modal version, batches are processed sequentially using callbacks. For the no-modal version, batches are sent in a loop.

### DIFF
--- a/tahrir/templates/_base.html
+++ b/tahrir/templates/_base.html
@@ -37,11 +37,28 @@
         }
         function claim_badges() {
           var urls = badge_urls();
-          {% if config['TAHRIR_OPENBADGES_MODAL'] %}
-          var lolback = function(errors, successes) {};
-          OpenBadges.issue(urls, lolback);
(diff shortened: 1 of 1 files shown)


gridhead (changes requested):
This is a change made against a legacy portion of the project codebase that has been selected for impending removal. Please consider introducing changes to the backend API and the frontend PWA instead of focusing on the server-side views and Jina-based frontend.

Closing this pull request without merging.

PR #887, opened 2026-03-29
Secure GET /api/invitations endpoint

### LINKED ISSUE #885: Invitations endpoint missing authentication (IDOR)
The endpoint added in c2c395f was intentionally left incomplete per its TODO comments.

The fix is straightforward since OIDC is already set up in the project (`claim_invitation` 
on the same blueprint already uses `g.oidc_user`):

- Require authentication (same as other endpoints)
- Drop `user_id` from the path and resolve the user from `g.oidc_user` instead

Without this, the endpoint is publicly accessible and any caller can read another 
user's invitations by guessing their user_id (IDOR).

Daniel-1600:
solves #885


gridhead (changes requested):
Please remove the tests.

gridhead (changes requested):
Also, please _sign_ your commit activities and _specify_ your AI tooling inside the commit messages, after you are done with resolving the merge conflicts.

Daniel-1600:
I squashed everything into one sign commit since the previous two commits weren't signed

tahrir/endpoints/admin/authorizations.py
```diff
@@ -64,8 +70,14 @@ def remove_authorization():
...
+        jsonify(
+            {
+                "message": (
+                    f"Badge {data.get(required_fields[0])!r} authorization revoked from "
+                    f"{data.get(required_fields[1])!r}"
+                )
+            }
+        ),
+        200,
+    )
```
sdglitched, line 77: I guess this refactoring was needed to secure the GET Invitations endpoint.

Let the linting and formater perform these changes. Please revert it.

tahrir/endpoints/admin/authorizations.py
```diff
@@ -31,11 +31,17 @@ def add_authorization():
...
+        jsonify(
+            {
+                "message": (
+                    f"Badge {data.get(required_fields[0])!r} authorized to "
+                    f"{data.get(required_fields[1])!r}"
+                )
+            }
+        ),
+        201,
+    )
```
sdglitched, line 41: I guess this refactoring was needed to secure the GET Invitations endpoint.

Let the linting and formater perform these changes. Please revert it.

tahrir/endpoints/assertions.py
```diff
@@ -11,7 +11,7 @@ def get_recent_assertions():
     begin = request.args.get("begin", 0, type=int)
     limit = min(request.args.get("limit", 40, type=int), 40)
 
-    assertions = list(g.tahrirdb.get_all_assertions(begin, limit))
+    assertions = list(g.tahrirdb.get_all_assertions().offset(begin).limit(limit).all())
```
sdglitched: Are you sure this is how [get_all_assertions](https://github.com/fedora-infra/tahrir-api/blob/43e1906e4872d8615f002d82a58e051a7c6fe5d7/tahrir_api/dbapi.py#L864-L880) works?

tahrir/endpoints/invitations.py
```diff
@@ -34,15 +33,19 @@ def claim_invitation(invitation_id: str):
     return jsonify({"message": f"You have earned badge {claim.badge_id!r}"})
 
 
-@bp.route("/api/invitations/<string:user_id>", methods=["GET"])
-def get_invitations(user_id: str):
+@csrf.exempt
```
sdglitched: CSRF (Cross-Site Request Forgery) protection is specifically designed to protect against unauthorized "unsafe" state-changing requests, such as POST, PUT, DELETE, and PATCH.

This is not needed for GET requests.

tahrir/utils/badge.py
```diff
@@ -39,7 +39,11 @@ def _badge_tags_list(badge):
 
 
 def badge_json_generator(badge, withasserts=True):
-    """Serialize a badge for the JSON API. When withasserts is False, returns compact data for listings."""
+    """
+    Serialize a badge for the JSON API.
+
+    When withasserts is False, returns compact data for listings.
+    """
```
sdglitched, line 46: I guess this refactoring was needed to secure the GET Invitations endpoint.

Let the linting and formater perform these changes. Please revert it.

tahrir/utils/badge.py
```diff
@@ -50,7 +54,7 @@ def badge_json_generator(badge, withasserts=True):
             "image": badge.image,
             "tags": _badge_tags_list(badge),
             "criteria": badge.criteria,
-            "rarity": raredata["badges"][badge.id]["rare"],
+            "rarity": raredata["badges"].get(badge.id, {}).get("rare", "common"),
```
sdglitched: This is a good practice, but as part of this project, a `Badge` will always have a rarity as it will be a database field itself.

So, having this change is not needed for now.

tahrir/utils/badge.py
```diff
@@ -103,7 +107,7 @@ def badge_json_generator(badge, withasserts=True):
         "tags": _badge_tags_list(badge),
         "issuer": badge.issuer.name,
         "criteria": badge.criteria,
-        "rarity": raredata["badges"][badge.id]["rare"],
+        "rarity": raredata["badges"].get(badge.id, {}).get("rare", "common"),
```
sdglitched: Same here.

tahrir/views/user.py
```diff
@@ -117,18 +115,11 @@ def _user_json_generator(person):
...
     for indx, item in enumerate(assertions):
         issued = {"issued": float(item.issued_on.strftime("%s"))}
         reason = {"reason": item.issued_for or None}
-        rarity = {"rarity": raredata["badges"][item.badge.id]["rare"]}
         badged = badge_json_generator(item.badge, withasserts=False)
-        serialized.append({**issued, **badged, **reason, **rarity})
+        serialized.append({**issued, **badged, **reason})
         for name in classified.keys():
             if name in item.badge.tags:
                 classified[name].append(indx)
```
sdglitched, line 134: This is a change made against the [legacy portion](https://github.com/fedora-infra/tahrir/wiki/Collaboration-Instructions-for-Outreachy-2026-Applicants#contributions) https://github.com/fedora-infra/tahrir/issues/904#issue-4183743250 of the project codebase that has been selected for impending removal.

Please revert.

sdglitched (changes requested):

tahrir/endpoints/assertions.py
Daniel-1600: Ahh I see my errror now . I was applying .offset().limit() outside the function without really looking into how pagination was already handled internally.

tahrir/utils/badge.py
Daniel-1600: ruff format doesn't auto-fix E501 in docstrings — it only formats code structure. The linter flags it but the formatter won't correct it, so a manual edit was the only way to resolve it. should I still revert it ?

Daniel-1600:
So I was looking at this endpoint and noticed something ,anyone can just pass in a user ID and get back someone else's data. 
there was no auth check, so you could just iterate through IDs and scrape all the user info. We should lock this down.
I noticed `/api/invitations` is already doing this the right way — it uses `g.oidc_user` to figure out who's asking. We should do the same thing here.

the  fix:
- Slap `@oidc.require_login` on the endpoint
- Drop the user_id from the URL parameter
- Just pull the user from `g.oidc_user.person` instead

That way the endpoint can only return data for the person making 
the request.

# Before (vulnerable):
@bp.route("/api/users/<string:user_id>", methods=["GET"])
def get_user_by_id(user_id: str):
    user = get_person(user_id)  # Anyone can guess IDs

# After (secure):
@bp.route("/api/users", methods=["GET"])
@oidc.require_login
def get_current_user():
    user = g.oidc_user.person  # Only authenticated user

I also tried getting someones data using their ID 

Screenshot From 2026-04-08 22-54-49

zsklav (commented):
The csrf part isnt needed, csrf is only for state-changing requests like POST/DELETE, GET requests are safe by defintion. Also the formatting changes in authorizations.py and badge.py arent related to the fix, maybe you should revert them.

Daniel-1600:
Thanks @Valkyriezz I will check on that

tahrir/utils/badge.py
sdglitched: Are you suggesting that running the format and lint using tox was failing? Can you share some evidence for that?

sdglitched:
@Daniel-1600, a PR should be atomic in nature, meaning that it only solves the issue it is created for. Other changes in files `authorization.py`, `assertions.py` and `badges.py` are not needed. Please revert them to the original state.

If the test fails in CI due to unrelated change, then we will check what is wrong.

Also, what were you trying to achieve by hitting the endpoint `/api/invitations/74629`?

Please provide the screenshot for the modified endpoint i.e., `/api/invitations/`

Daniel-1600:
> @Daniel-1600, a PR should be atomic in nature, meaning that it only solves the issue it is created for. Other changes in files `authorization.py`, `assertions.py` and `badges.py` are not needed. Please revert them to the original state.
> 
> If the test fails in CI due to unrelated change, then we will check what is wrong.
> 
> Also, what were you trying to achieve by hitting the endpoint `/api/invitations/74629`?
> 
> Please provide the screenshot for the modified endpoint i.e., `/api/invitations/`

My bad i shared the wrong screenshot i was just tetsing if the implimentation is also present in the that endpoint too.

tahrir/endpoints/invitations.py
```diff
@@ -34,15 +33,18 @@ def claim_invitation(invitation_id: str):
...
+@bp.route("/api/invitations", methods=["GET"])
+def get_invitations():
     """Endpoint to list all invitations created by a given user."""
 
-    # TODO - Modify the endpoint to require authentication
-    # TODO - Remove the user_id path parameter requirement
-    person = get_person(user_id)
+    email = g.oidc_user.email
+    if not email:
+        return abort(401, "Unauthorized")
```
mscherer, line 43: The relationship between the condition and the error is not obvious, could this be explained a bit more ?

Daniel-1600: > The relationship between the condition and the error is not obvious, could this be explained a bit more ?

so basically when g.oidc_user.email reads the token from fedora auth system and finds no email is there (probably cause the token is broken or not authenticated) it automatically rejects and returns 401

PR #894, opened 2026-03-30
Fix/setup test suite

System625:
## Summary
  Set up the test suite infrastructure so subsequent PRs can add tests      
  without re-initialising fixtures each time.

  - Add `tests/conftest.py` with `app`, `client`, `request_context`, and    
    `dummy_issuer` fixtures
  - Reset the dogpile cache singleton in the `app` fixture to prevent       
    `RegionAlreadyConfigured` errors across test files

  Assisted-by: Claude (Anthropic)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,61 @@
+import os
+
+import pytest
+from tahrir_api.utils import get_db_manager_from_uri
+
+from tahrir.app import create_app
+from tahrir.cache import cache
+from tahrir.database import db
+
+
+@pytest.fixture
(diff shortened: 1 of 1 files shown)


sdglitched:
@System625 I appreciate your efforts, but as the revamp is underway and the initial setup of test suite is considered an important task, we should wait for the maintainers to initialise it and then the mentees and other contributors can take it forward.

gridhead (changes requested):
Please wait for the maintainers to establish the testing configuration before adding your changes on top of it, instead of creating one by yourself from the ground up.

PR #902, opened 2026-04-01
include indeterminate spinner for showing progress in lookup text fields

### LINKED ISSUE #769: Include indeterminate spinner for showing progress on lookup text fields
Include indeterminate spinner for showing progress on lookup text fields

Daniel-1600:
closes #769

description 

https://github.com/user-attachments/assets/1a438db4-b9f0-42dc-9218-2ef4ab8631b6

### DIFF
--- a/frontend/src/crud/users/update.jsx
+++ b/frontend/src/crud/users/update.jsx
@@ -2,12 +2,13 @@ import { useState } from "react";
 import { Button, Card, Col, Dropdown, FloatingLabel, Form, Image, Row } from "react-bootstrap";
 import { useDispatch } from "react-redux";
 
+import { LookupSpinner } from "../../components/LookupSpinner.jsx";
 import {
   useLookupIdentityQuery,
   useToggleIdentityOptOutMutation,
   useUpdationIdentityMutation,
 } from "../../features/call.js";
-import { useLoadingState } from "../../features/hooks.js";
+import { useLoadingState, useMinFetching } from "../../features/hooks.js";
 import { showBaseNote } from "../../features/part.js";
 import { formatTime, portraitProvider } from "../../features/util.js";
 
@@ -31,9 +32,11 @@ export default function UserUpdateForm() {
   const [userLookup, setUserLookup] = useState("");
   const [userDropdownShow, setUserDropdownShow] = useState(false);
 
-  const { data: searchResults } = useLookupIdentityQuery(userLookup, {
+  const { data: searchResults, isFetching: isUserFetching } = useLookupIdentityQuery(userLookup, {
     skip: userLookup.length < 4,
   });
+
+  const showUserSpinner = useMinFetching(isUserFetching);
 
   useLoadingState(isUpdating, isToggling);
 
@@ -176,6 +179,7 @@ export default function UserUpdateForm() {
                   autoComplete="off"
                 />
               </FloatingLabel>
+              {showUserSpinner && <LookupSpinner />}
               {userLookup.length >= 4 &&
                 searchResults &&
                 searchResults.users &&
--- a/frontend/src/features/hooks.js
+++ b/frontend/src/features/hooks.js
@@ -1,4 +1,4 @@
-import { useEffect } from "react";
+import { useEffect, useRef, useState } from "react";
 import { useDispatch } from "react-redux";
 
 import { hideLoad, showLoad } from "./part.js";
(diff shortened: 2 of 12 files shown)


gridhead (changes requested):
Please _sign_ your commits and _specify_ what AI tooling or model you have made use of within the commit messages!

gridhead (changes requested):
Also, please limit the length of your commit messages to under fifty characters.

frontend/src/crud/users/update.jsx
```diff
@@ -176,6 +178,12 @@ export default function UserUpdateForm() {
                   autoComplete="off"
                 />
               </FloatingLabel>
+              {showUserSpinner && (
+                <div className="lookup-spinner">
+                  <div className="spoke" /><div className="spoke" /><div className="spoke" /><div className="spoke" />
+                  <div className="spoke" /><div className="spoke" /><div className="spoke" /><div className="spoke" />
+                </div>
```
gridhead: If I could point to something that is most definitely NOT generated by AI, I would point towards these lines of codebase changes. I see what you're trying to do here, and I notice your innocent mistake here. You would want to move these things away from here to make a component out of this so that it could be reproduced elsewhere, too. Take, for instance, there are lookups inside the administration page that you would want to take a look at next.

gridhead: Try something in the lines of


```
function LookupSpinner() {
  return (
    <div className="lookup-spinner">
      {Array.from({ length: 8 }, (_, i) => <div key={i} className="spoke" />)}
    </div>
  );
}
```


```suggestion
                <LookupSpinner />
```

I don't know if it works, though.

Give it a try.

Let me know if you need more help with this.

frontend/src/styles/main.css
```diff
@@ -257,3 +257,41 @@ a.vibe-border {
 [data-bs-theme="dark"] a.vibe-border:hover {
   background-color: rgba(224, 224, 224, 0.05) !important;
 }
+
+/* Lookup field spinner - pinwheel style */
+.lookup-spinner {
+  position: absolute;
```
gridhead, line 263: Why is it like this here? What would happen if it were to be changed to the following?

```suggestion
  position: relative;
```

frontend/src/styles/main.css
```diff
@@ -257,3 +257,41 @@ a.vibe-border {
...
+  transform: translateY(-50%);
+  z-index: 4;
+  pointer-events: none;
+  width: 16px;
+  height: 16px;
+  animation: lookup-spin 0.8s steps(8) infinite;
+}
+
+.lookup-spinner .spoke {
+  position: absolute;
```
gridhead, line 275: Why is it like this here? What would happen if it were to be changed to the following?

```suggestion
  position: relative;
```

frontend/src/styles/main.css
```diff
@@ -257,3 +257,41 @@ a.vibe-border {
...
+}
+
+.lookup-spinner .spoke:nth-child(1) { transform: rotate(0deg);   opacity: 1.0; }
+.lookup-spinner .spoke:nth-child(2) { transform: rotate(45deg);  opacity: 0.875; }
+.lookup-spinner .spoke:nth-child(3) { transform: rotate(90deg);  opacity: 0.75; }
+.lookup-spinner .spoke:nth-child(4) { transform: rotate(135deg); opacity: 0.625; }
+.lookup-spinner .spoke:nth-child(5) { transform: rotate(180deg); opacity: 0.5; }
+.lookup-spinner .spoke:nth-child(6) { transform: rotate(225deg); opacity: 0.375; }
+.lookup-spinner .spoke:nth-child(7) { transform: rotate(270deg); opacity: 0.25; }
+.lookup-spinner .spoke:nth-child(8) { transform: rotate(315deg); opacity: 0.125; }
```
gridhead: Here too. I see what you are trying to do here and I appreciate you trying to do it the hard way manually rather than relying on some clearly polished AI-generated code, but please consider modularising it to ensure that it can be easily reproduced else where without the shenanigans that we have worry about over here.

frontend/src/features/hooks.js
```diff
@@ -14,3 +14,40 @@ export function useLoadingState(isLoading, isUpdating) {
...
+        setVisible(false);
+        timerRef.current = null;
+      }, remaining);
+    }
+    return () => {
+      if (timerRef.current) {
+        clearTimeout(timerRef.current);
+      }
+    };
+  }, [isFetching]);
```
gridhead, line 53: The `useEffect` dependency array only lists `isFetching`, but the effect body reads `visible` and `minMs`. This is intentional (the effect should only fire on fetch transitions, not on visibility changes), but React's `exhaustive-deps` rule will flag it. I will recommend consider adding either those two as well or explaining why they are not there because they will sure as heck be flagged by the linter on a run.

gridhead (changes requested):
Please make the following suggested changes.

frontend/src/crud/users/update.jsx
Daniel-1600: lemme try and see

frontend/src/styles/main.css
Daniel-1600: I used absolute to overlap the surrounding elements, I think for relative it would kinda push the surrounding elements cause it will work like a div . That was my approach

frontend/src/styles/main.css
Daniel-1600: it wouldn't retain it's position like this 

https://github.com/user-attachments/assets/0897efcc-e6f2-45f6-9242-eba4f3be0d86

frontend/src/features/hooks.js
Daniel-1600: I left out minMs cause its constant so my reasoning was we don't need to keep on watching it  and for visible i also left out cause it will create an endless loop since the it will always be changing like when a user types 4 character and it becomes true (cause that's when we start fetching the data ) then when i delete the characters becomes false   that will keep on triggering the effect.

gridhead (changes requested):
Apart from the actual change itself (which looks pretty good), the logistics around the change are rather underwhelming.

Image

Please read up on how to properly sign your commit activities inside the commit message, as I notice two different names here.

Your commit messages (or at least the first one) are also a whole lot longer than the maximum recommended limit of 50 characters.

The [AI-assisted Contribution Policy](https://docs.fedoraproject.org/en-US/council/policy/ai-contribution-policy/) also requires you to have an `Assisted-by: ` commit activity message trailer, not a custom format.

The worst part is that these are the teachings that have been repeatedly talked about, and yet I notice frequent lapses everywhere.

I am going to make changes to yours, not because your efforts show promise, but at least, your AI-assisted codebase changes do.

Daniel-1600:
@gridhead  I want to acknowledge the issues raised in this review. The commit messages were too long, my name was inconsistent across commits, and the AI disclosure was not in the correct Assisted-by: format as required by the contribution policy. I will make sure future PRs follow the correct standards before submission

PR #905, opened 2026-04-01
Refactor: Hydration error warning caused by nested hyperlink HTML ele…

### LINKED ISSUE #903: [Bug]: Hydration error warning caused by nested <a> HTML elements.
Hydration error warning pops up every time the fedora badge home page is loaded.
This  can lead to unexpected behaviors, such as broken HTML elements, missing or doubled event listeners, and even unhandled state changes.

Image

thedunncodes:
## Summary

- Hydration error warning pops up every time the home page is loaded, This is caused by nesting an hyperlink element e.g `<a>` in another. Many reusable components in the modern packages often have different components implementing an HTML hyperlink element in them. When these components are not used properly it could result into an hydration error. 

- In the tahrir project, this error occurs at `tahrir/frontend/src/routes/homepage.jsx` where we have a `<Link>` component nested in a `<VertItem>` component which also contains a `<Link>` component.

Fixes #903 

## Changes

- The `<VertItem>` component has been provisioned with an `onClick` option which can be used in place of the `link` option in this case. 

- I created a `handleAwardsClicks` function to satisy the `onClick` event in the `<VertItem>` component. Within the function, navigation is properly handled by the `useNavigation`hook from react router.

Screenshot From 2026-04-01 04-14-01

### DIFF
--- a/frontend/public/docs/addendum.md
+++ b/frontend/public/docs/addendum.md
@@ -1,0 +1,75 @@
+# Fedora Badges
+
+[Fedora Badges](https://badges.fedoraproject.org) is a fun website built to recognize contributors to the [Fedora Project](https://fedoraproject.org), help new and existing Fedora contributors find different ways to get involved, and encourage the improvement of Fedora's infrastructure.
+
+![Fedora Badges](https://badges.fedoraproject.org/pngs/badges_fan.png)
+
+## How does Badges work?
+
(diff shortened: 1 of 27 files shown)


thedunncodes:
This Pull request can't be merged due to the resolved conflicts, unneccessary files are being pushed

Issue #904, opened 2026-04-01
Remove legacy Jinja2 frontend and associated dead code


gridhead:
The project has been migrating from a legacy Jinja2 template-based frontend to a modern React PWA (`frontend/`).

The legacy frontend is still present in the codebase and has been causing confusion; contributors have been submitting patches against dead or soon-to-be-removed code instead of focusing on the React frontend.

This umbrella issue tracks the removal of all legacy frontend code and associated infrastructure. Please create one pull request per functionality listed below so that each removal can be reviewed and merged independently.

Each of the following is a self-contained removal/cleanup activity. Please pick one, create a branch, remove the code, verify nothing in the React frontend or API endpoints breaks, and open a PR referencing this issue.

1. Remove legacy Jinja2 templates
   - **Directory:** `tahrir/templates/`
   - All `.html` template files (`index.html`, `user.html`, `badge.html`, `badge_full.html`, `leaderboard.html`, `explore.html`, `explore_badges.html`, `report.html`, `about.html`, `diff.html`, `admin.html`, `builder.html`, `thingiview.html`, `assertion_widget.html`, `tags.html`, `error.html`, `_base.html`, `_badge-base.html`, `_functions.html`)
   - Includes the dead `test.html` template that is not referenced by any view

2. Remove legacy view functions that render Jinja2 templates
   - **Directory:** `tahrir/views/`
   - Views that render legacy templates: `index`, `about`, `user`, `diff`, `badge`, `badge_full`, `leaderboard`, `report`, `explore`, `explore_badges`, `admin`, `builder`, `thingiview`, `assertion_widget`, `tags`
   - **Do not remove** the API endpoint functions in `tahrir/endpoints/` — those serve the React frontend

3. Remove legacy static assets (CSS)
   - **Directory:** `tahrir/static/css/`
   - Files: `unsemantic-grid-mobile.css`, `unsemantic-grid-responsive.css`, `unsemantic-grid-responsive-tablet.css`, `tahrir.css`, `monokai.css`
   - These serve the old Jinja2-based HTML frontend only

4. Remove legacy static assets (JavaScript)
   - **Directory:** `tahrir/static/js/`
   - Files: `jquery-3.3.1.min.js`, `social.js`, `favico-0.3.4.min.js`, `builder.js`
   - These are used exclusively by the legacy Jinja2 templates

5. Remove the 3D STL badge viewer
   - **Directory:** `tahrir/static/js/thingiview/`
[...]

PR #908, opened 2026-04-01
cleanup: remove dead code in tahrir/utils/avatar.py

Linsmed:
Part of remove legacy Jinja2 frontend and associated dead code #904
I cleaned up dead code in tahrir/utils/avatar.py

tahrir/utils/avatar.py
Lines 41–50 are unreachable due to an early return on line 39
 the dead code and the associated TODO comment on lines 36–3 were removed

### DIFF
--- a/tahrir/utils/avatar.py
+++ b/tahrir/utils/avatar.py
@@ -5,14 +5,6 @@ from flask import current_app
 
 from tahrir.cache import cache
 
-
-libravatar = None
-try:
-    import libravatar
-except ImportError:
-    pass
-
-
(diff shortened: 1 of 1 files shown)


gridhead (changes requested):
Closing this in favour of #906, not only because #906 was opened earlier, but also since there is a lot of content left out here unaddressed that has been taken care of over there. That is, of course, not to mention that the commits have not been signed here as well. Please look into problems related to inconsistent spacing, commit message length limits, unnecessarily left trailing whitespaces and removal of obsolete functionalities.

PR #909, opened 2026-04-01
cleanup: remove legacy CSS files and references

Linsmed:
Part of #904 — Legacy frontend removal

What's removed:
All 5 files in tahrir/static/css/ and the 4 corresponding <link> tags in tahrir/templates/_base.html.

The key safety argument is straightforward: CSS files are purely presentational, consumed only by browsers rendering legacy HTML pages. They have no connection to the React frontend or to any API endpoint. Removing them and their <link> references is a clean, zero-risk operation for the new frontend.

PR #910, opened 2026-04-01
cleanup: remove legacy JavaScript files and inline script blocks

Linsmed:
Part of #904 — Legacy frontend removal.

What's removed:

4 JS files from tahrir/static/js/ (jQuery, social.js, favico, builder.js)
3 <script src> tags + 2 inline <script> blocks in _base.html that depended on them
1 <script src> tag in builder.html


gridhead (changes requested):
Could you please combine this with #909 and also address the changes that I have suggested there? I am going ahead with closing this pull request.

PR #913, opened 2026-04-01
Remove legacy Jinja2 JavaScript static assets

payalrvs3:
Remove JS files exclusively used by the old Jinja2-based HTML frontend:

1. jquery-3.3.1.min.js
2. social.js
3. favico-0.3.4.min.js
4. builder.js

These are only referenced from tahrir/templates/ and have no connection to the React PWA frontend or any API endpoint.

Template references are intentionally left unchanged, as the templates themselves are scheduled for removal in a separate task(task1).

Part of #904"


gridhead (changes requested):
TLDR. Please target the `swatantrya` branch and not the `develop` branch.

The [legacy UI](https://github.com/fedora-infra/tahrir/wiki/Collaboration-Instructions-for-Outreachy-2026-Applicants#contributions) will no longer be worked on. Please check the `frontend` directory to know about the newer improvements we are introducing to the development stack. As this change would eventually be disposed of, I am electing not to accept this change. Please make changes to the ReactJS-based frontend revamp directory rather than the Jinja one.

PR #919, opened 2026-04-01
Hydration error warning caused by nested hyperlink HTML elements

### LINKED ISSUE #903: [Bug]: Hydration error warning caused by nested <a> HTML elements.

thedunncodes:
## Summary

- Hydration error warning pops up every time the home page is loaded, This is caused by nesting an hyperlink element e.g `<a>` in another. Many reusable components in the modern packages often have different components implementing an HTML hyperlink element in them. When these components are not used properly it could result into an hydration error. 

- In the tahrir project, this error occurs at `tahrir/frontend/src/routes/homepage.jsx` where we have a `<Link>` component nested under a `<VertItem>` component which also contains a `<Link>` component.

Fixes #903 
**No AI tools used**

## Changes

- The `<VertItem>` component has been provisioned with an `onClick` option which can be used in place of the `link` option in this case. 

- I created a `handleAwardsClicks` function to satisy the `onClick` event in the `<VertItem>` component. Within the function, navigation is properly handled by the `useNavigation` hook from react router.

Screenshot From 2026-04-01 04-14-01

### DIFF
--- a/frontend/src/routes/homepage.jsx
+++ b/frontend/src/routes/homepage.jsx
@@ -40,6 +40,14 @@ export default function Homepage() {
     }
   };
 
+  // Using onClick instead of link to avoid nested <a> tags (hydration error).
+  // Trade-off: right-click "Open in new tab" will not work on award items.
+  const handleAwardsClick = (e, path) => {
+    if(!e.target.closest('a')) {
+      navigate(path);
+    }
+  };
+
   // Show or Hide LoadNote
   useEffect(() => {
     if (isLoading) {
@@ -195,7 +203,7 @@ export default function Homepage() {
                     </>
                   }
                   shot={item.person?.mail ? portraitProvider(item.person.mail, 45) : null}
-                  link={`/identity/${item.person?.nickname || ""}`}
+                  onClick={(e) => {handleAwardsClick(e, `/identity/${item.person?.nickname || ""}`)}}
                 />
               ))}
             </ListGroup>


gridhead (approved):
While I am not happy about you not signing the commit activities and specifying the AI tooling (if any was used), I noticed your insights in finding out the issue #903 and your efforts on fixing it. I made an exception this time around by making amending changes on top of your changes instead of sending it around for another round of reviews, because I see that your work here shows promise.

Please take a look at the changes that I have made, not only to the codebase but to the way I have squashed your commits, reworded the messages and disclosed the usage of assistive AI-based toolings. Use this opportunity to learn from this and then find more avenues of contribution both within this project's codebase as well as in the wider Fedora Project community.

thedunncodes:
@gridhead Thank you very much for your insight on this.
I was aware of the trade-off, because making that fix would turn the components into a button like component where links can't be previewed or copied. My mistake was not notifying that and I'll ensure i keep that mind for future contributions.
As for the signed commits, I didn't know initially but after you raised the concern on the group, I have taken note of that and my last commit on this PR was signed.
I didn't use any AI tools though, so i thought i didn't need to specify i used one. Although i reviewed other people's PR and i loved the way they drafted their comments for their PRs so I copied that, even though their drafts might be done by AI. I look forward to learning more about this in today's meeting.

Issue #929, opened 2026-04-05
Refactor badge search logic into reusable function


Daniel-1600:
Currently, the badge search logic is directly inside the endpoint.
We should move it into a reusable function to improve maintainability and allow other parts of the code to reuse it

PR #932, opened 2026-04-05
Refactor search badges to use DB-level pagination

Daniel-1600:
solves #929

This is entirely dependent to  https://github.com/fedora-infra/tahrir-api/issues/267#issue-4206077174 pr


Daniel-1600:
I had closed cause I thought the logic was wrong,and logistics of it

PR #938, opened 2026-04-05
fix: undefined variable in user_team_json

### LINKED ISSUE #937: NameError crash in user_team_json due to undefined variable `user`
The `user_team_json` view in `tahrir/views/user.py` references an undefined variable `user` in its null check: get_person() returns its result into person, but the null check uses user which was never defined anywhere in this function. This means:

- Every call to /user/<user_id>/team/<int:team_id>/json crashes immediately with a 'user' is not defined error
- The 404 guard never runs
- Valid users never get their team data
- Invalid users never get a proper error 

### Fix
 Replace `user` with `person` in the null check.

Marvinrose:
fixes #937 

Fix: replace `user` with `person` in the null check.

### DIFF
--- a/tahrir/views/user.py
+++ b/tahrir/views/user.py
@@ -179,7 +179,7 @@ def user_team_json(user_id, team_id):
     person = get_person(user_id)
     team = g.tahrirdb.get_team(team_id=team_id)
 
-    if not user:
+    if not person:
         return {"error": "No such user exists."}, 404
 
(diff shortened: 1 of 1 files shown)


gridhead (approved):
You need not specify the `Assisted-by: ` commit activity message trailer if you have not used any AI tooling.

With that being said, this change looks good to me but this legacy code is leaving us with the issue ticket #904.

PR #940, opened 2026-04-06
fix: correct decorator order on invitation endpoints

### LINKED ISSUE #941: Fix decorator order on invitation endpoints
While going through the code, I noticed that @bp.route was placed at the bottom of the decorator stack in both tahrir/endpoints/invitations.py and tahrir/endpoints/admin/invitations.py. Since Flask registers whichever function @bp.route directly wraps at import time, having it below @csrf.exempt and @oidc.require_login means the raw unprotected function gets registered, completely bypassing login and CSRF protection.

payalrvs3:
While going through the code, I noticed that `@bp.route` was placed at the bottom of the decorator stack in both `tahrir/endpoints/invitations.py` and `tahrir/endpoints/admin/invitations.py`. Since Flask registers whichever function `@bp.route` directly wraps at import time, having it below `@csrf.exempt` and `@oidc.require_login` meant the raw unprotected function was being registered instead of the wrapped one, completely bypassing the login and CSRF protection. Moving `@bp.route` to the outermost position fixes this so Flask registers the fully protected version. Ran the full test suite before and after and all 4 tests pass with no regressions.

Fixes #941

### DIFF
--- a/tahrir/endpoints/admin/invitations.py
+++ b/tahrir/endpoints/admin/invitations.py
@@ -7,10 +7,10 @@ from ...utils.user import get_person, require_admin
 from . import blueprint as bp
 
 
+@bp.route("/api/admin/invitations", methods=["POST"])
 @csrf.exempt
 @oidc.require_login
 @require_admin
-@bp.route("/api/admin/invitations", methods=["POST"])
 def add_invitations():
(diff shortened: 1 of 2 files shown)


gridhead (approved):
LGTM.

Merging.

PR #943, opened 2026-04-06
fix: add missing session.commit() in opt_out

zsklav:
This PR fixes an issue where the user `opt-out` endpoint updates the opt_out field only in memory without persisting it to the database. As a result, the change is lost after the request completes. The fix adds a missing `g.tahrirdb.session.commit()` call to ensure the updated value is properly saved.


gridhead (approved):
LGTM.

Merging.

PR #946, opened 2026-04-06
fix: use timezone-aware datetimes in API endpoints

Linsmed:
Affected folders and lines
endpoints/invitations.py line 25: claim.expires_on < datetime.now()
endpoints/invitations.py line 32: add_assertion(..., datetime.now())
endpoints/admin/assertions.py line 35: datetime.fromtimestamp(issued_on)
endpoints/admin/invitations.py lines 24, 33:


zsklav (commented):
Good catch on the naive datetimes! One thing tho, datetime.now(tz=timezone.utc).replace(tzinfo=None) in invitations.py makes it tz-aware then strips it right away which kinda defeats the purpose. 
Also the poetry.lock diff is huge and unrealted to the fix, maybe a mistake while staging the files?

mscherer:
Give how the function is used, wouldn't it break the use case where a admin set a expiration with a specific timezone in the timestamp sent in the API call ?

PR #951, opened 2026-04-07
feat: add REST API endpoints for  awarding and invitations

### LINKED ISSUE #949: Missing REST API endpoints for badge awarding & invitations
The old Jinja views in `views/badge.py` let authorized users award badges and create invitations, but they use form data and redirects. The React frontend needs JSON APIs instead. Admin endpoints exist for this at `/api/admin/assertions` , but those are admin-only. There's no JSON API for regular authorized users to award or invite. 

This issue proposes adding new REST API endpoints in `endpoints/badges.py` to support delegated badge awarding and invitation, following the pattern used by existing admin badge endpoints.

zsklav:
Adds POST endpoints for badge awarding and invitations to support JSON-based requests from the React frontend. Includes validation and authorization checks, tested locally in Vagrant.

Fixes #949


tahrir/endpoints/badges.py
```diff
@@ -143,3 +146,72 @@ def search_badges_by_string(search_string: str):
...
+
+    result = g.tahrirdb.add_invitation(
+        badge.id,
+        expires_on=expires_on,
+        created_by_email=agent.email,
+    )
+    if not result:
+        return abort(400, "Failed to create invitation")
+
+    return jsonify({"message": f"Invitation created for badge {badge_id!r}"}), 201
```
Marvinrose: Suggestion: Since this endpoint creates an invitation, would it be possible to return atleast the invitation ID (or the invitation payload) in the response? It would make the endpoint easier to consume and more REST-friendly.

tahrir/endpoints/badges.py
```diff
@@ -143,3 +146,72 @@ def search_badges_by_string(search_string: str):
...
+@bp.route("/api/badges/<badge_id>/award", methods=["POST"])
+@csrf.exempt
+@oidc.require_login
+def award(badge_id):
+    badge = g.tahrirdb.get_badge(badge_id)
+    if not badge:
+        return abort(404, f"No such badge {badge_id!r}")
+    # Check if the logged-in user is authorized to award this badge
+    if not badge.authorized(g.oidc_user.person):
+        return abort(403, f"Unauthorized for {badge_id!r}")
```
Daniel-1600: just a small correction i have seen you have used 403 and "unauthorized" at the same time which refer to two different things.  403 (forbidden) and 401 (unauthorized) . Consider changing it to something like f"Forbidden: not permitted to award {badge_id!r}" to match what 403 actually means.

tahrir/endpoints/badges.py
```diff
@@ -143,3 +146,72 @@ def search_badges_by_string(search_string: str):
...
+@oidc.require_login
+def invite(badge_id):
+    agent = g.oidc_user.person
+
+    badge = g.tahrirdb.get_badge(badge_id)
+    if not badge:
+        return abort(404, f"No such badge {badge_id!r}")
+    # Check if the logged-in user is authorized to invite for this badge
+    if not badge.authorized(agent):
+        return abort(403, f"Unauthorized for {badge_id!r}")
```
Daniel-1600: this too

zsklav: Good catch you're right  403 means Forbidden, not Unauthorized. I ll update the message to `f"Forbidden: not permitted to award {badge_id!r}" `(and the same for the invite endpoint). Thanks for reviewing

tahrir/endpoints/badges.py
zsklav: That's a great point i ll  update the response to include the invitation ID and details so the frontend can use it directly thanks for the suggestion

zsklav:
Good catch you're right 403 means Forbidden, not Unauthorized. I ll update the message to f"Forbidden: not permitted to award {badge_id!r}" (and the same for the invite endpoint). Thanks for reviewing

zsklav:
That's a great point. I'll update the response to include the invitation ID and details so the frontend can use it directly. Thanks for the suggestion!

zsklav:
While making the changes to the API endpoints I also noticed that the older form-based views in `tahrir/views/badge.py`(coz i took reference from there) had the same incorrect `'Unauthorized'` wording on their 403 responses. Since those views serve the same award and invite functionality, I went ahead and updated them too to keep the error messaging consistent across both the new REST API and the legacy form based endpoints

PR #912, opened 2026-04-01
Remove legacy Jinja2 templates

zsklav:
## Summary
- Delete all 21 Jinja2 template files from `tahrir/templates/`
`.html` template files (`index.html`, `user.html`, `badge.html`, `badge_full.html`, `leaderboard.html`, `explore.html`, `explore_badges.html`, `report.html`, `about.html`, `diff.html`, `admin.html`, `builder.html`, `thingiview.html`, `assertion_widget.html`, `tags.html`, `error.html`, `_base.html`, `_badge-base.html`, `_functions.html`,`test.html`) 
- These templates are part of the legacy frontend being replaced by React
- No API endpoints or React frontend code references these templates

Fixes part of #904 (Task 1)

PR #953, opened 2026-04-08
cleanup: remove dead template infrastructure from app.py

### LINKED ISSUE #952: Dead template infrastructure remains in app.py after Jinja2 template removal
while going through the codebase ,I noticed that while the Jinja2 templates were removed in e0372f2, however,  the supporting infrastructure that fed data to those templates was left behind. This includes a context processor, two Jinja2 filter registrations, two utility modules, a config value, and a data directory (all now dead code.)

Linsmed:
Fixes #952
The Jinja2 templates were removed in e0372f2 but the infrastructure that supported them was left behind. This PR removes all of it.

what I deleted:

utils/templates.py
utils/docs.py 
tahrir/sitedocs/ — which contains(about.rst, footer.rst)

The lines of code that I removed from existing files are:
app.py — 3 imports and the "Templates" section 
utils/avatar.py — as_avatar() function 
utils/date_time.py — relative_time() function + import math + import dateutil.relativedelta (only caller was the deleted filter)
defaults.py — TAHRIR_SITEDOCS_SUBDIR config (only reader was deleted docs.py)


gridhead (changes requested):
Before I begin with a formal round of reviews, could you please consistently _sign_ your commit activities and _specify_ your AI tooling inside your commit messages? Also, please ensure that your commit message lengths are under the 50-character limit and squash the uselessly repeatedly mentioned commits.

Linsmed:
> Before I begin with a formal round of reviews, could you please consistently _sign_ your commit activities and _specify_ your AI tooling inside your commit messages? Also, please ensure that your commit message lengths are under the 50-character limit and squash the uselessly repeatedly mentioned commits.
> 
>  

Okay sir

thedunncodes:
@Linsmed I don't think the sitedocs files should be removed though. It's meant to inform a new contributor about tahrir. You should look at how to update it.

The tahrir folder serves as backend for the frontend folder, both within Tahrir. There should be a documentation showing their relation to new contributors.

Fnd out if the  functions within `utils/doc.py` can help in improving the site docs, if not the reasons for deleting them would be valid

zsklav (commented):
Hey nice cleanup! I actually worked on the Jinja2 template removal in #912 so this caught my eye. One thing tho, thedunncodes has a point about the sitedocs, maybe the content should be migrated somwhere into the React side instead of just deleting it? Also in defaults.py the comment changed from # to ### which looks accidental. And probably should try grepping for relative_time and as_avatar before removing just to make sure nothng else imports them. Looks good otherwise!

PR #875, opened 2026-03-28
fix: Improve CSV badge awarding with validation and error reporting

Marvinrose:
## Fix: CSV badge awarding with validation and error reporting

Fixes #348 

### Problem
The `award_from_csv` endpoint existed to allow admins to award badges in bulk via CSV upload, but had a TODO requesting validation and feedback that was never implemented.

The original code had file validation issues, bytes read as strings, silent failures with no admin feedback..

### Fix
To fix this issue, all i did was:

- Validate the file is present and has a filename before processing
- Decode each line from bytes to string explicitly using UTF-8
- Normalise column order: both `email,badge_id` and `badge_id,email` are accepted
- Validate badge existence before attempting to award
- Auto-create users who don't exist, preserving original behaviour and report which users were created
- Track and report already-awarded badges instead of silently skipping
- Flash a complete summary after processing, covering all outcomes

PR #871, opened 2026-03-27
Refactor user search with database-level pagination

Daniel-1600:
closes #864

PR #958, opened 2026-04-09
feat: Add CSV badge awarding API endpoint

### LINKED ISSUE #348: RFE: csv submission of badge awardees
I have a list of ~30 or so Flock speakers I have to award the speaker badge to. It takes about 2-3 minutes per speaker to award (type in speaker name, if the FAS acct is wrong gives me a 404 error and I have to dig around, I have to hit back and fresh reload the page to try again, once it's awarded it brings me to an RSS feed and I have to hit back and reload again)

Life would be a lot easier if I could just upload a csv of fas accounts to award the badge, and it spit out which accounts weren't found.

Marvinrose:
Fixes #348 

### Problem
The CSV badge awarding had multiple issues: no file validation for missing or empty file uploads, no feedback to the admin on what happened etc.. This fix is implemented in the correct location in the new API-based architecture.
The was initially implemented in #875 but was closed because it was fixed in a portion of the codebase that was selected for removal.

### Changes
-  Added POST /api/admin/badges/award/csv endpoint in tahrir/endpoints/admin/badges.py
-  Returns structured JSON reporting successful awards, auto-created users, already awarded badges, badges not found, and malformed lines
- Added tests covering all edge cases in tests/test_csv_badge_awarding.py

### Approach
Unlike the legacy flash-message implementation, this endpoint returns structured JSON consistent with the existing admin API endpoints..

Screenshot 2026-04-09 174058


thedunncodes:
The test environment set up by the AI tool is off, let me list some of the errors 

- There isn't a need for a new sqlalchemy connection and where it's connecting to is also wrong. This would have worked using `TahrirDatabase` from the tahrir-api
- The wrong test file is being referenced
- The other test functions are okay though, but it wouldn't scale for testing the whole tahrir endpoints and utils

I think there's also an endpoint that does something like this, the only fix needed was to add a validation and flash messages for failed or successful awards. It's in the `tahrir/views/admin.py`'

zsklav (commented):
Hey I think 'awards[email] = badge_id' will silently overwrite if the same email shows up with different badges, maybe a list of tuples can be used instead?
And is the @csrf.exempt intentional? And thedunncodes had a good point about using TahrirDatabase in the tests insted of raw sqlalchemy

tahrir/endpoints/admin/badges.py
```diff
@@ -81,3 +81,77 @@ def delete_badge(badge_id):
...
+            continue
+        if not line:
+            continue
+        values = line.split(",")
+        if len(values) != 2:
+            malformed_lines.append(line)
+            continue
+        value_a = values[0].strip()
+        value_b = values[1].strip()
+        if "@" in value_a:
```
mscherer: This seems fragile, there is no restriction on badge name as far as I know, so it could have a '@' in its name.

tahrir/endpoints/admin/badges.py
```diff
@@ -81,3 +81,77 @@ def delete_badge(badge_id):
...
+
+    for raw_line in csv_file:
+        try:
+            line = raw_line.decode("utf-8").strip()
+        except UnicodeDecodeError:
+            malformed_lines.append(repr(raw_line))
+            continue
+        if not line:
+            continue
+        values = line.split(",")
```
mscherer: any reason to not use the standard library here (eg, the `csv` [module](https://docs.python.org/3/library/csv.html)) ?

tahrir/endpoints/admin/badges.py
```diff
@@ -81,3 +81,77 @@ def delete_badge(badge_id):
...
+        awards[email] = badge_id
+
+    if not awards and not malformed_lines:
+        return abort(400, "The CSV file was empty")
+
+    for email, badge_id in awards.items():
+        if not g.tahrirdb.badge_exists(badge_id):
+            not_found_badges.append(badge_id)
+            continue
+        if not g.tahrirdb.person_exists(email=email):
```
mscherer, line 157: This is fragile, since a typo in the CSV file will create a user instead of showing a error.

tahrir/endpoints/admin/badges.py
```diff
@@ -81,3 +81,77 @@ def delete_badge(badge_id):
...
+            g.tahrirdb.add_person(email)
+            auto_created_persons.append(email)
+        if not g.tahrirdb.assertion_exists(badge_id, email):
+            g.tahrirdb.add_assertion(badge_id, email, None)
+            successful_awards += 1
+        else:
+            already_awarded.append(f"{email} ({badge_id})")
+
+    return jsonify({
+        "message": f"Successfully awarded {successful_awards} badge(s).",
```
mscherer: Isn't it a problem that even a incorrect CSV file will say it is a success ?

tests/test_csv_badge_awarding.py
```diff
@@ -0,0 +1,151 @@
...
+        response = client.post(
+            "/api/admin/badges/award/csv",
+            data=data,
+            content_type="multipart/form-data"
+        )
+        result = response.get_json()
+        assert result["successful_awards"] == 1
+
+
+def test_csv_already_awarded(client):
```
mscherer: This only work if the test are run in the same order, and not if someone run only 1 test or if the order change later. This should be fixed.

Marvinrose:
> The test environment set up by the AI tool is off, let me list some of the errors
> 
> * There isn't a need for a new sqlalchemy connection and where it's connecting to is also wrong. This would have worked using `TahrirDatabase` from the tahrir-api
> * The wrong test file is being referenced
> * The other test functions are okay though, but it wouldn't scale for testing the whole tahrir endpoints and utils
> 
> I think there's also an endpoint that does something like this, the only fix needed was to add a validation and flash messages for failed or successful awards. It's in the `tahrir/views/admin.py`'

Hi @thedunncodes 
Thanks for the review! 
- I didn't use a raw SQLAlchemy, rather, I used a MagicMock to mock the database entirely which is actually valid for testing? @gridhead 
- What do you mean by the wrong test file is being referenced please?
- Your point about it not scaling is irrelevant to this PR since I'm only testing one endpoint.
- The endpoint you're trying to reference `tahrir/views/admin.py` is a part of the legacy portion of the project codebase that has been selected for impending removal. See #904

Marvinrose:
> Hey I think 'awards[email] = badge_id' will silently overwrite if the same email shows up with different badges, maybe a list of tuples can be used instead? And is the @csrf.exempt intentional? And thedunncodes had a good point about using TahrirDatabase in the tests insted of raw sqlalchemy

Hi @Valkyriezz 
Thanks for the review!
Thanks for pointing out that bug, I've fixed it with a list of tuples in my latest commit. And yes the @csrf.exempt is intentional. All the other admin API endpoints have it too because the frontend sends JSON requests without CSRF tokens. 
I'm just following the codebase structure.

tahrir/endpoints/admin/badges.py
Marvinrose: > any reason to not use the standard library here (eg, the `csv` [module](https://docs.python.org/3/library/csv.html)) ?

You're right about using a csv module, its a better practice, I've updated my latest commit to use it.. Thank you

tahrir/endpoints/admin/badges.py
Marvinrose: > Isn't it a problem that even a incorrect CSV file will say it is a success ?

You're right, this is misleading.. I've updated it in my latest commit.

tahrir/endpoints/admin/badges.py
Marvinrose: > This is fragile, since a typo in the CSV file will create a user instead of showing a error.

You're right! I've updated it in my latest commit..

tahrir/endpoints/admin/badges.py
Marvinrose: > This seems fragile, there is no restriction on badge name as far as I know, so it could have a '@' in its name.

 I've updated it in my latest commit.. thanks

thedunncodes:
> > The test environment set up by the AI tool is off, let me list some of the errors
> > 
> > * There isn't a need for a new sqlalchemy connection and where it's connecting to is also wrong. This would have worked using `TahrirDatabase` from the tahrir-api
> > * The wrong test file is being referenced
> > * The other test functions are okay though, but it wouldn't scale for testing the whole tahrir endpoints and utils
> > 
> > I think there's also an endpoint that does something like this, the only fix needed was to add a validation and flash messages for failed or successful awards. It's in the `tahrir/views/admin.py`'
> 
> Hi @thedunncodes Thanks for the review!
> 
>     * I didn't use a raw SQLAlchemy, rather, I used a MagicMock to mock the database entirely which is actually valid for testing? @gridhead
> 
>     * What do you mean by the wrong test file is being referenced please?
> 
>     * Your point about it not scaling is irrelevant to this PR since I'm only testing one endpoint.
> 
>     * The endpoint you're trying to reference `tahrir/views/admin.py`' is a part of the legacy portion of the project codebase that has been selected for impending removal. See [Remove legacy Jinja2 frontend and associated dead code #904](https://github.com/fedora-infra/tahrir/issues/904)


Glad to see you have fixed the issues, for your questions:

- Alright, that's great
- Pytest has a format for naming test files, since you are testing `badges.py` I was expecting a name like `test_badges.py`. However, I'm aware the test files can still run but then for a collaborator who is new to the repo, it might throw the individual off a bit. It's maintaining structure for easy access.
- I'm aware you are testing one file which is absolutely fine, but the connections are just limited to that file even if you are testing other endpoints within `badges.py`. For example, the MagicMock connection you made can be made into a module inside `tests/conftest.py` which would be made available to any other test file under the `tests/` dir. This would prevent making new connections in every test file and benefit other files under `tahrir` yet to be tested.
- The endpoint is what's left after @Valkyriezz  cleaned up the legacy view functions that were obsolete in that dir, @gridhead assigned it to her under a different task which she also completed under issue #904, but there wasn't any merged request that says view functions have been removed to easily identify it. This is the kind of issue I'm addressing in the way you referenced your test file lol. 

But great work!

Marvinrose:
Also, I think I should remove the tests because I saw in PR #887 and #871 where @gridhead and @sdglitched asked Daniel to remove the tests in his PR but I’ll leave the final decision to them.

tahrir/endpoints/admin/badges.py
```diff
@@ -81,3 +84,101 @@ def delete_badge(badge_id):
...
+
+@bp.route("/api/admin/badges/award/csv", methods=["POST"])
+@csrf.exempt
+@oidc.require_login
+@require_admin
+def award_badges_from_csv():
+    """Endpoint to award badges to multiple users from a CSV file.
+
+    Expected CSV format (one per line):
+    email,badge_id
```
mscherer, line 97: This PR should support only 1 format. Supporting both, as pointed in my previous comment is increasing complexity for no good reason.

mscherer:
I think the API (and the returned json) should be simplified by making sure that all badges can be awarded, or show a error and award 0 badge if there is any error in the file. This would simplify the code and usage (since a API user do not need to check 4 keys in the result of the function).

PR #960, opened 2026-04-09
Implement a Clear button in the search bar to improve user experience

### LINKED ISSUE #959: Implement a clear button in the search bar to improve UX
A clear button should appear when users input their search strings, this improves the user experience.

Image

thedunncodes:
## Overview

Most browsers don't support the use of 'type=search' in input types which often brings an 'x' button to help users clear their inputs. FIrefox definitely doesn't support the search type, I also tried on a chrome browser it didn't show up then I found out a css style which targets the elements could help show it on chrome, but this wouldn't be a permanent solution. 
A clear button was added to the search bar to improve the UX, and is being handled by a `handleClear` function

close #959 
**No AI tools used**

### Before 
Screenshot From 2026-04-09 15-42-42

### After

https://github.com/user-attachments/assets/5f5ded1c-1c63-46c4-9e1f-a8e641be77dc

### DIFF
--- a/frontend/src/components/discover.jsx
+++ b/frontend/src/components/discover.jsx
@@ -10,6 +10,7 @@ import { LookupSpinner } from "./LookupSpinner.jsx";
 export default function Discover() {
   const [findText, makeFindText] = useState("");
   const [dropSeen, makeDropSeen] = useState(false);
+  const [clearSeen, makeClearSeen] = useState(false);
   const dropDown = useRef(null);
   const navigate = useNavigate();
   const doLookup = findText.length >= 4;
@@ -41,6 +42,7 @@ export default function Discover() {
     const text = e.target.value;
     makeFindText(text);
     makeDropSeen(text.trim().length >= 4);
+    makeClearSeen(text.trim().length >= 1);
   };
 
   const handleResult = () => {
@@ -62,6 +64,12 @@ export default function Discover() {
     }
   };
 
+  const handleClear = () => {
+    makeFindText("");
+    makeClearSeen(false);
+    makeDropSeen(false);
+  };
+
   const hasResults = searchResults && (searchResults.badges?.length > 0 || searchResults.users?.length > 0);
 
   return (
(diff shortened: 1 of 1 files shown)


zsklav (changes requested):
Hey @thedunncodes ,
It is indeed a nice addition but you have mistakenly pushed a commit that is not relevant to this PR. 
Another thing is handleClear doesn't call makeDropSeen(false) so the dropdown might stay visible after clearing the input.

thedunncodes:
makeDropSeen(false) gets handled by the handleChange function being used to manage state change in the search box input field. 

But yes, when I make a PR, previous commits made to another one automatically joins it, how can i stop that @Valkyriezz

zsklav:
that happens when u branch off from a branch that already has those commits. always create your new branch from the base branch (swatantrya in this case). for now u can cherry-pick just your relevant commit onto a fresh branch from swatantrya and force push

zsklav:
To fix this PR, run:


git fetch upstream
git checkout -b clear_discover_field_fix upstream/swatantrya
git cherry-pick 741c792
git push origin clear_discover_field_fix:clear_discover_field --force
Replace upstream with whatever you named the fedora-infra/tahrir remote (check with git remote -v).

This creates a clean branch from swatantrya, picks only your clear button commit, and force pushes to update this PR. The unrelated commit will be gone.

For future PRs, always branch off from swatantrya directly:


git checkout swatantrya
git pull upstream swatantrya
git checkout -b your-new-branch

additionally 
f you're using any AI tooling, make sure to add attribution in your commit message like:
Authored-by: <tool-name> <noreply@example.com>

thedunncodes:
@Valkyriezz I have fixed it, thanks for the help

frontend/src/components/discover.jsx
```diff
@@ -62,6 +64,11 @@ export default function Discover() {
     }
   };
 
+  const handleClear = () => {
+    makeFindText('')
+    makeClearSeen(false);
```
Daniel-1600, line 69: why didn't you include makeDropSeen(false) if I may ask

thedunncodes: makeDropSeen(false) gets handled by the handleChange function being used to manage state change in the search box input field.

Daniel-1600:
tested it locally and looks great. Thanks @thedunncodes 

https://github.com/user-attachments/assets/26597553-873d-4bf9-ad54-5df816a402f2

gridhead (approved):
LGTM 🌮

PR #963, opened 2026-04-12
fix: paginate get_all_assertions in Python

### LINKED ISSUE #962: get_all_assertions() called with invalid arguments causing TypeError
While setting up the project locally and testing the API, I got a TypeError on the `/api/assertions` endpoint.

## Error
The endpoint was failing because `get_all_assertions()` in tahrir-api takes no arguments, but in `tahrir/endpoints/assertions.py` it was being called with `begin` and `limit`:

```python
assertions = list(g.tahrirdb.get_all_assertions(begin, limit))
```

This caused the recent awards section on the homepage to fail to load.

## Fix
Fetch all assertions first, then slice in Python, same pattern already used in `get_assertions_by_badge`

payalrvs3:
Closes #962

While setting up the project locally, the homepage was showing "Resource unavailable" and the browser console showed a 500 error on `/api/assertions?begin=0`.

Turns out `get_all_assertions()` takes no arguments, but `begin` and `limit` were being passed directly to it. Fixed by fetching 
all assertions first and then slicing in Python, which is the same approach already used in `get_assertions_by_badge`.

```python
all_assertions = list(g.tahrirdb.get_all_assertions())
assertions = all_assertions[begin:begin + limit]
```

**Before:** 500 Internal Server Error, homepage fails to load
Screenshot 2026-04-12 114226

**After:** Recent awards load correctly
Screenshot 2026-04-12 162110

### DIFF
--- a/tahrir/endpoints/assertions.py
+++ b/tahrir/endpoints/assertions.py
@@ -12,7 +12,7 @@ def get_recent_assertions():
     limit = min(request.args.get("limit", 40, type=int), 40)
 
     all_assertions = list(g.tahrirdb.get_all_assertions())
-    assertions = all_assertions[begin:begin + limit]
+    assertions = all_assertions[begin : begin + limit]
     if not assertions:
         return jsonify([])
 
--- a/tahrir/endpoints/assertions.py
+++ b/tahrir/endpoints/assertions.py
@@ -11,7 +11,8 @@ def get_recent_assertions():
     begin = request.args.get("begin", 0, type=int)
     limit = min(request.args.get("limit", 40, type=int), 40)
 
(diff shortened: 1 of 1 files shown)


sdglitched:
@payalrvs3, it seems you are using older version of [tahrir-api](https://github.com/fedora-infra/tahrir-api) or facing some kind of setup issues.
image

The current version of [tahrir-api](https://github.com/fedora-infra/tahrir-api), i.e., 1.4.3, takes `begin` and `limit` as its arguments
image

payalrvs3:
@sdglitched Thank you for the review! I understand now. I was using an older version of tahrir-api. The original code is correct for version 1.4.3. Closing this PR and the related issue.

PR #965, opened 2026-04-12
fix: check user auth before fetching API

### LINKED ISSUE #964: expired token check happens after fetch in loadUserData
In `frontend/src/features/auth.js`, the `loadUserData` function fetches `/api/users/` using `user.access_token` before checking if the user is null or expired.

## Problem
If `user` is null, accessing `user.access_token` throws a TypeError immediately. If the token is expired, the fetch runs with a stale token 
causing a 401 response.

payalrvs3:
Closes #964

While reviewing the authentication flow, `loadUserData` in `frontend/src/features/auth.js` was fetching `/api/users/` using `user.access_token` before checking if the user was null or expired.

If `user` is null, accessing `user.access_token` throws a TypeError. If the token is expired, the fetch runs with a stale token causing a 401 response.

Fixed by moving the null/expired check before the fetch.

## Changes
- Only `frontend/src/features/auth.js` is changed

### DIFF
--- a/frontend/src/features/auth.js
+++ b/frontend/src/features/auth.js
@@ -6,6 +6,9 @@ import { API_BASE_URL } from "./call.js";
 
 export const loadUserData = createAsyncThunk("auth/loadUserData", async () => {
   const user = await userManager.getUser();
+  if (!user || user.expired) {
+    return null;
+  }
   const response = await fetch(API_BASE_URL + "/api/users/", {
     method: "POST",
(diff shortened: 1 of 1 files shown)


Daniel-1600 (commented):
looks good to me thanks @payalrvs3

gridhead (approved):
LGTM 🌮 

Thanks for working on this, @payalrvs3!

PR #966, opened 2026-05-08
Include deployment controls for Fedora Infrastructure

gridhead:
Include deployment controls for Fedora Infrastructure

@abompard, is this correct?

### DIFF
--- a/.s2i/bin/assemble
+++ b/.s2i/bin/assemble
@@ -12,3 +12,12 @@ pip install . --no-deps
 # set permissions for any installed artifacts
 fix-permissions /opt/app-root -P
 echo "---> Installing application and dependencies ..."
+
+# Build the frontend
+cd frontend
+npm ci
+
+if [ "$FEDORA_ENV" == "staging" ]; then
+    NPM_EXTRA_ARGS="-- --mode staging"
+fi
+npm run build ${NPM_EXTRA_ARGS}
--- a/tahrir/app.py
+++ b/tahrir/app.py
@@ -21,7 +21,7 @@ from tahrir.utils.avatar import as_avatar
 from tahrir.utils.date_time import relative_time
 from tahrir.utils.templates import templates_context
 from tahrir.utils.user import on_authorized
-from tahrir.views import add_static_view, internal_server_error, page_not_found
+from tahrir.views import add_frontend_view, add_static_view, internal_server_error, page_not_found
 from tahrir.views import blueprint as root_bp
 
 
@@ -136,6 +136,10 @@ def create_app(config=None):
     add_static_view(app, app.config["TAHRIR_PNGS_PATH"], prefix="/pngs", endpoint="pngs")
     if app.config.get("TAHRIR_STLS_PATH"):
         add_static_view(app, app.config["TAHRIR_STLS_PATH"], prefix="/stls", endpoint="stls")
+
+    # Frontend SPA
+    if frontend_path := app.config.get("TAHRIR_FRONTEND_PATH"):
+        add_frontend_view(app, frontend_path)
 
     # CLI
     app.cli.add_command(tahrir_cli)
--- a/tahrir/defaults.py
+++ b/tahrir/defaults.py
@@ -67,3 +67,5 @@ CACHE = {
 }
 
 FLASK_ADMIN_FLUID_LAYOUT = True
+
+TAHRIR_FRONTEND_PATH = os.path.join(BASE_PATH, "..", "frontend", "dist")
--- a/tahrir/views/__init__.py
+++ b/tahrir/views/__init__.py
@@ -1,3 +1,4 @@
+import werkzeug.exceptions
 from flask import Blueprint, Flask, jsonify
 from flask.helpers import send_from_directory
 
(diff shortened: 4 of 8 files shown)


.s2i/start.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+exec gunicorn \
```
abompard: I thought that gunicorn would be detected and used by the default s2i `run` script when the `APP_MODULE` env var is set. Why did you need to create this shell script?

tahrir/views/__init__.py
```diff
@@ -19,3 +21,13 @@ def _view(filename):
         return send_from_directory(directory, filename, max_age=max_age)
 
     app.add_url_rule(f"{prefix}/<path:filename>", endpoint=endpoint, view_func=_view)
+
+
+def add_frontend_view(app: Flask, directory: str):
+    def _serve_frontend(path=""):
```
abompard: It would be nice to have unit tests for this. And also a test that runs `_server_frontend("/etc/passwd")` and makes sure the output does not contain the file (because `os.path.join("/a", "/b")` is `/b`).

abompard (changes requested):

tahrir/views/__init__.py
```diff
@@ -19,3 +22,13 @@ def _view(filename):
...
+        filepath = safe_join(directory, path)
+        if path and filepath and os.path.isfile(filepath):
```
gridhead: To respond to https://github.com/fedora-infra/tahrir/pull/966/changes#r3217396215, how about this change?

abompard: Actually, [the doc for send_from_directory()](https://werkzeug.palletsprojects.com/en/stable/utils/#werkzeug.utils.send_from_directory) already uses `safe_join` to ensure the file is inside the given directory. So I suppose we could instead simplify this function and do something like:
```py
try:
    return send_from_directory(directory, path)
except werkzeug.exceptions.NotFound:
    return send_from_directory(directory, "index.html")
```
What do you think?

tahrir/views/__init__.py
```diff
@@ -1,5 +1,8 @@
+import os
+
 from flask import Blueprint, Flask, jsonify
 from flask.helpers import send_from_directory
+from werkzeug.security import safe_join
```
abompard, line 4: If you decide to follow the above suggestion, this can be removed. Otherwise, note that `safe_join` was moved from `werkzeug.security` to `werkzeug.utils` in Werkzeug 2.2 and removed from `werkzeug.security` in Werkzeug 3.0. Since this project uses Flask 3.x (which requires Werkzeug 3.x), we can update the import.

tahrir/app.py
```diff
@@ -137,6 +137,9 @@ def create_app(config=None):
     if app.config.get("TAHRIR_STLS_PATH"):
         add_static_view(app, app.config["TAHRIR_STLS_PATH"], prefix="/stls", endpoint="stls")
 
+    # Frontend SPA
+    add_frontend_view(app, app.config["TAHRIR_FRONTEND_PATH"])
+
```
abompard, line 143: In development, `frontend/dist/` does not exist (the Vite dev server runs separately on port 5173), so any path that does not match a registered route will hit this and fail with a 500. Consider gating it:
```python
if frontend_path := app.config.get("TAHRIR_FRONTEND_PATH"):
    add_frontend_view(app, frontend_path)
```

abompard (changes requested):

gridhead:
How about this change?

.s2i/bin/assemble
```diff
@@ -12,3 +12,12 @@ pip install . --no-deps
 # set permissions for any installed artifacts
 fix-permissions /opt/app-root -P
 echo "---> Installing application and dependencies ..."
+
+# Build the frontend
+cd frontend
+npm install
```
abompard: Prefer `npm ci` over `npm install` here. In a build/CI context `npm ci` is faster for clean installs, won't silently mutate `package-lock.json`, and fails if the lockfile is out of sync with `package.json`.

tahrir/views/__init__.py
```diff
@@ -19,3 +20,13 @@ def _view(filename):
         return send_from_directory(directory, filename, max_age=max_age)
 
     app.add_url_rule(f"{prefix}/<path:filename>", endpoint=endpoint, view_func=_view)
+
+
+def add_frontend_view(app: Flask, directory: str):
+    def _serve_frontend(path=""):
```
abompard: This is minor but I don't see any case where the `path` argument would be left out, you can remove the `=""`.

tahrir/defaults.py
```diff
@@ -67,3 +67,5 @@
 }
 
 FLASK_ADMIN_FLUID_LAYOUT = True
+
+TAHRIR_FRONTEND_PATH = os.path.join(BASE_PATH, "..", "frontend", "dist")
```
abompard, line 71: Please also add `TAHRIR_FRONTEND_PATH = ""` to `devel/ansible/roles/dev/files/tahrir.cfg` for the development environment.

abompard (changes requested):
Only minor stuff!

gridhead:
Ok how about now?

abompard (approved):
LGTM!

PR #968, opened 2026-05-20
Add unit tests for `_serve_frontend` path traversal safety

### LINKED ISSUE #967: Include unit tests for the `_serve_frontend` function path
Include unit tests for the `_serve_frontend` function path

`add_frontend_view` in `tahrir/views/__init__.py` uses `os.path.join(directory, path)` to check whether a requested file exists. When `path` is an absolute path (e.g., `/etc/passwd`), `os.path.join` silently discards the `directory` prefix — `os.path.join("/a", "/b")` returns `/b`. This means the `os.path.isfile` check probes an attacker-controlled path.

While Flask's `send_from_directory` has its own traversal protection internally, the `os.path.isfile` gate leaks information about arbitrary file existence on the host. We need to replace the `os.path.join` with `werkzeug.security.safe_join`, which returns `None` for absolute or ..-based traversal paths (that do not belong to the current subdirectory).

sheikhlimon:
Add unit tests for `_serve_frontend` in `tahrir/views/__init__.py`.

- test SPA fallback to `index.html` on unknown paths
- test that `/etc/passwd` does not leak host file contents via path traversal

Closes #967


tests/test_views.py
```diff
@@ -0,0 +1,36 @@
...
+        response = client.get("/some/spa/route")
+    assert response.status_code == 200
+    assert b"index" in response.data
+
+
+def test_path_traversal_absolute(app):
+    with app.test_client() as client:
+        response = client.get("/etc/passwd")
+    assert response.status_code == 200
+    assert b"index" in response.data
```
gridhead: Can these be parametrized? Yes, they can be.

tests/test_views.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@pytest.fixture
+def app(frontend_dir):
+    app = Flask(__name__)
+    app.config["TESTING"] = True
+    add_frontend_view(app, frontend_dir)
+    return app
+
+
```
gridhead, line 28: Also, include relative traversal vectors too, not just absolute ones.

Like `/../../../etc/passwd` and `/%2e%2e/%2e%2e/etc/passwd`, for starters.

gridhead, line 28: You are just checking for the exclusivity of the sensitive directories.

To add to that, please verify for the exclusion of sensitive contents.

Like responses that qualify `assert b"root:" not in response.data`.

We not only want to check for the fallback but also for the filtering.

tests/test_views.py
```diff
@@ -0,0 +1,36 @@
...
+import pytest
+from flask import Flask
+
+from tahrir.views import add_frontend_view
+
+
+@pytest.fixture
+def frontend_dir(tmp_path):
+    (tmp_path / "index.html").write_text("<html>index</html>")
+    return str(tmp_path)
```
gridhead: Add more parameter checks here for fixture generated files.

Checking for `index.html` feels rather done for the sake of it.

gridhead (changes requested):
Thanks for the draft.

Please take another look.

tests/test_views.py
sheikhlimon: of course

tests/test_views.py
sheikhlimon: I was mainly focused on the absolute-path case from #967 initially, but adding relative + encoded traversal cases seems worthwhile too.

sheikhlimon: Agreed on the negative assertion as well

tests/test_views.py
sheikhlimon: I originally focused more on the fallback behavior here, but agreed that checking actual fixture-generated files would make the tests more representative.

sheikhlimon:
Thanks for the review! I've pushed the suggested updates. Let me know if there’s anything else you'd like adjusted.

PR #971, opened 2026-05-28
Switch from `session-based OIDC` to `Authentication Bearer`

gridhead:
Switch from `session-based OIDC` to `Authentication Bearer`

Oooh hooO


gridhead:
Okay here goes nothing....

Issue #977, opened 2026-06-11
fix: TAHRIR_USE_OPENID_EMAIL not respected consistently in auth utils


Daniel-1600:
While going through the auth utils I noticed that `create_person()` already handles `TAHRIR_USE_OPENID_EMAIL` correctly, but three other places in the same file `User.email`, `get_awarded_assertions()` and `_populate_access_user()` still hardcode the domain email and never check this flag.  The fix would be applying the same pattern from `create_person()` to those three places.

PR #978, opened 2026-06-11
fix: respect TAHRIR_USE_OPENID_EMAIL in auth

Daniel-1600:
solve issue #977


Daniel-1600:
closing this issue since the after close review the oidc is only needed on `create_person()` since I saw at the database level it uses constructed email even on `create_person()`  i saw its stores constructed email

PR #985, opened 2026-07-21
Include contrast page for friendly comparison

### LINKED ISSUE #882: feat: Convert diff page from Jinja2 template to React component
As part of the frontend revamp on the swatantrya branch, the diff.html Jinja2 template needs to be converted to a React component.

The diff page compares two users' badges side by side showing:
- Badges unique to user A
- Badges unique to user B  
- Badges shared by both users
- Rank and percentile comparison stats

This involves:
- Adding a /json/diff/<id_a>/<id_b> API endpoint in tahrir/views/user.py
- Adding retrieveDiff query in frontend/src/features/call.js
- Creating frontend/src/routes/diff.jsx React component
- Registering /diff/:id_a/:id_b route in frontend/src/main.jsx

gridhead:
Include contrast page for friendly comparison

Fixes #882

### DIFF
--- a/frontend/src/components/accoitem.jsx
+++ b/frontend/src/components/accoitem.jsx
@@ -4,7 +4,7 @@ import { Link } from "react-router";
 
 import { obtainRarityBack, obtainRarityEdge, obtainRarityText, relativeImageUrl } from "../features/util.js";
 
-export default function AccoItem({ iden, name, body, foot, shot, rare }) {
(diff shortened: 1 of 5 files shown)


gridhead:
This is what it looks like right now.

You gotta login first.

image

Then click on that **Compare** button.

image

There's the comparison screen.

You cannot compare people without login.

You cannot compare people if you are not one of them too.

PR #1012, opened 2026-07-27
Restore rankings endpoint under REST conventions

gridhead:
Restore rankings endpoint under REST conventions


gridhead:
I might be stupid. This was already merged with #1011.

PR #1003, opened 2026-07-25
Add sticky sidebar with hidden overflow scroll on desktop

sdglitched:
## Summary
- Apply `sticky-lg-top` and `align-self-lg-start` to left-column containers across all two-column pages so sidebar cards remain visible while scrolling
- Uses responsive classes so sticky behavior does not activate on mobile where the columns stack vertically
- Excluded the `badges` detail page where the sidebar content (badge image, description, tier, status, first/last awarded, pagination) exceeds viewport height

## Affected pages
- itempage (covers assembly, livelist, deadlist, recently)
- homepage, addendum, governor
- identity, userpast, contrast, campaign
- category, rarities, rankings, findpage

Fixes: #992 

## Some Examples
image
image
image
image
image
image

I've tested it on different screen sizes, but it may need further testing to ensure no edge cases with overflow.

PR #1015, opened 2026-08-10
Include navigation for search result sections

gridhead:
Include navigation for search result sections


gridhead:
Helps with long navigations

- When the **"Badges" button** is clicked

image

- When the **"Users" button** is clicked

image

But honestly, consider precise keyword searching

Builds atop of @sdglitched's excellent work on #1003

PR #1016, opened 2026-08-10
Include navigation for governor entities sections

gridhead:
Include navigation for governor entities sections


gridhead:
Helps with long navigations

- When the **"Assertions" button** is clicked

image

- When the **"Invitations" button** is clicked

image

- When the **"Authorizations" button** is clicked

image

- When the **"Badges" button** is clicked

image

- When the **"Users" button** is clicked

image

Desktop users please rejoice

Uses #1003

